PR #7, opened 2019-04-08
37161 precompute materialinput for perf

labkey-matthewb:


flow/src/org/labkey/flow/query/FlowSchema.java
```diff
@@ -1100,9 +1100,14 @@ public DataType getDataType()
         public ColumnInfo addMaterialInputColumn(String alias, SamplesSchema schema, String inputRole, ExpSampleSet sampleSet)
         {
             checkLocked();
+            /*
```
labkey-kevink: delete me

PR #14, opened 2019-04-22
Issue 37278: IllegalStateException in print view of table with XML-customized buttons

labkey-jeckels:
Remove immutable default ButtonBars, unscoped to a controller. Update the small number of places I could find that rely on those default button, mainly very old code in experiment, user, and mothership UI.

Remove display mode from DisplayElement. Only configure the buttons and columns that are desired for the type of view that is being used (insert, update, grid, details, etc)


labkey-jeckels:
Since there was a minor SVN change (committed there this morning with revision 63443) I did a remote run. Found and fixed one problem in the BVTs. DailyC, arbitrarily chosen, seemed OK as well with one perhaps intermittent failure.

Thus, there may be some small issues, but I don't expect widespread problems.

PR #15, opened 2019-04-23
More targeted NPE prevention. `null` experiment is sometimes expected

labkey-tchad:


labkey-jeckels:
@labkey-kevink looks like you approved from a non-LabKey account, so this is still waiting review from an account with commit access.

PR #23, opened 2019-05-15
Fix AMT export when starting from nested protein grid view

labkey-jeckels:
Problem was that we were ending up trying to filter by RowId using a Protein Group id on a peptide column. Fixed by setting up the filter before forcing the column list to the AMT expected columns.

Includes some cleanup refactoring, including collapsing two abstract classes into one, adding @Override annotations, and other simplifications that were possible after removing the legacy MS2 views earlier.


labkey-tchad:
@labkey-jeckels The `MS2Test` AMT export checks are passing but we're getting some unexpected behavior when executing a custom query.
```
java.lang.AssertionError: Text 'Could not resolve IndistinguishableCollectionId column' was not present the correct number of times expected:<13> but was:<100>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.labkey.test.WebDriverWrapper.assertTextPresent(WebDriverWrapper.java:1562)
	at org.labkey.test.tests.ms2.MS2Test.queryValidationTest(MS2Test.java:1206)
	at org.labkey.test.tests.ms2.MS2Test.verifyMS2(MS2Test.java:86)
	at org.labkey.test.tests.ms2.AbstractMS2ImportTest.testSteps(AbstractMS2ImportTest.java:44)
```
https://teamcity.labkey.org/viewLog.html?buildId=748713#testNameId-3074434568554713685

labkey-jeckels:
@labkey-adam @labkey-tchad @labkey-sweta Looks like the subsequent failure was likely introduced by the immutable tableinfo changes.

Regardless of the cause, the commit I just pushed into this branch should fix the problem. Requesting review just of that followup change.

PR #34, opened 2019-06-05
Issue 37604: link flow samples to fcs files on sample import

labkey-kevink:
Issue 37604: link flow samples to fcs files on sample import
- call SampleSet.onSamplesUpdated to notify FlowProtocolImplementation


labkey-matthewb:
Needs regression test

PR #46, opened 2019-07-11
Fix crawler error setting bad proxy URL on Mascot settings from Admin…

labkey-jeckels:
… Console

Misc code cleanup, all auto-generated refactors compliments of IntelliJ


labkey-jeckels:
@labkey-ankurj I know you haven't dealt with this code, but the functional change here is just to demote a warning from an ERROR to an INFO condition in the log file on line 148. The rest is automated refactor code cleanup

PR #50, opened 2019-07-26
FB: Microarray Module Archiving

RosalineP:


microarray/src/org/labkey/microarray/MicroarrayFolderType.java
```diff
@@ -35,13 +35,10 @@ public MicroarrayFolderType(MicroarrayModule module)
         super("Microarray",
```
labkey-jeckels: I don't think this folder type is useful anymore. I suggest just deleting it. The assay folder type provides almost the same setup.

microarray/src/org/labkey/microarray/MicroarrayModule.java
```diff
@@ -68,8 +59,6 @@
 
     public static final AssayDataType MAGE_ML_INPUT_TYPE =
```
labkey-jeckels: Can any of these other AssayDataTypes go away too?

microarray/src/org/labkey/microarray/MicroarrayRunType.java
```diff
@@ -20,12 +20,12 @@
 import org.labkey.api.data.ButtonBar;
 import org.labkey.api.data.ContainerFilter;
 import org.labkey.api.exp.ExperimentRunType;
-import org.labkey.api.exp.Lsid;
+//import org.labkey.api.exp.Lsid;
```
labkey-jeckels: Just delete these imports if they're not needed, no reason to keep them around as comments.

microarray/src/org/labkey/microarray/assay/MicroarrayAssayProvider.java
```diff
@@ -39,9 +39,7 @@
 import org.labkey.api.study.assay.AssayDataType;
```
labkey-jeckels: I think this class should be deleted.

.gitattributes
```diff
@@ -681,7 +678,6 @@ microarray/src/org/labkey/microarray/MicroarrayRunUploadForm.java -text
 microarray/src/org/labkey/microarray/MicroarrayStatisticsView.java -text
 microarray/src/org/labkey/microarray/MicroarrayUploadWizardAction.java -text
 microarray/src/org/labkey/microarray/PendingMageMLFilesView.java -text
-microarray/src/org/labkey/microarray/affy/AffymetrixAssayProvider.java -text
```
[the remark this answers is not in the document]
labkey-klum: It won't do any harm to leave it in there but our recommendation is to remove the corresponding entry when deleting a file to simplify the .gitattributes file

microarray/resources/schemas/dbscripts/postgresql/microarray-19.20-19.21.sql
```diff
@@ -1,22 +1,18 @@
-/*
```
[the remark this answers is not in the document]
labkey-klum: Adam and I noticed this as well... I'm hoping this is just git/github being dumb but would be good for Rosaline to verify

.gitattributes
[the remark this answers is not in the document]
labkey-adam: I wrote a small local tool that parses .gitattributes and logs any file path that doesn't exist, which would indicate a delete, move, or rename. I've used it to keep the platform file in sync with my assay refactor (and other changes that pre-date my refactor). If we care about maintaining these files then I'd recommend that we build this into the server itself as a junit test or an action (dev mode only when sources are available, of course).

My only hesitation: despite my last few commits keeping .gitattributes 100% in sync with my file moves, the GitHub web view continues to show whole-file diffs :(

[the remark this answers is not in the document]
labkey-adam: Feel free to open an issue or stick a small story on the backlog, if you want me to commit this code.

microarray/resources/schemas/dbscripts/postgresql/microarray-19.20-19.21.sql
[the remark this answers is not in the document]
RosalineP: Not sure how FileStatus.java is being pulled in here, since I simply deleted it...
I remember creating the upgrade script by copying the previous upgrade script file contents and changing the  few lines necessary.

microarray/src/org/labkey/microarray/MicroarrayFolderType.java
RosalineP: Deleted!

microarray/src/org/labkey/microarray/MicroarrayRunType.java
RosalineP: Oops, I intended to. Thanks for this catch.

microarray/src/org/labkey/microarray/query/MicroarrayUserSchema.java
```diff
@@ -155,10 +153,6 @@ public TableInfo getAnnotationSetTable()
         return getTable(TABLE_FEATURE_ANNOTATION_SET);
     }
 
-    public TableInfo getAnnotationTable()
```
labkey-kevink: I think we still need the FeatureAnnotation table.

microarray/webapp/WEB-INF/microarray/microarrayContext.xml
```diff
@@ -2,51 +2,4 @@
 <beans xmlns="http://www.springframework.org/schema/beans"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-2.0.xsd">
-
-    <bean id="microarrayPipelineRegistry" class="org.labkey.api.pipeline.TaskPipelineRegistrar">
```
labkey-kevink: I think we can just delete this file.

microarray/src/org/labkey/microarray/MicroarrayManager.java
```diff
@@ -79,12 +79,6 @@ public static MicroarrayManager get()
     }
 
 
-    private static TableInfo getAnnotationSetQueryTableInfo(User user, Container container)
```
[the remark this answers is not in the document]
RosalineP: I also thought this was odd. I've done a little bit of hand-testing on this, and ran both the test files, and the feature annotation set functionality seems to be in place. Perhaps there's a use case I'm missing while testing?

microarray/src/org/labkey/microarray/query/MicroarrayUserSchema.java
RosalineP: As an addon to my comment above, if you think it'd be safer to keep it without finding the missing use case, I'd be happy to do that

microarray/src/org/labkey/microarray/MicroarrayModule.java
```diff
@@ -188,9 +142,6 @@ protected void startupAfterSpringConfig(ModuleContext moduleContext)
             return Collections.emptySet();
         });
 
-        SearchService ss = SearchService.get();
-        if (null != ss)
-            ss.addDocumentParser(new MageMLDocumentParser());
```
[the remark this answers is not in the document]
RosalineP: Done!

microarray/src/org/labkey/microarray/assay/MicroarrayAssayProvider.java
RosalineP: Done!

microarray/src/org/labkey/microarray/controllers/MicroarrayController.java
```diff
@@ -165,21 +93,5 @@ public NavTree appendNavTrail(NavTree root)
...
     }
-
-
-    @RequiresPermission(ReadPermission.class)
-    public class MatrixQueryAction extends SimpleViewAction
-    {
-        @Override
-        public ModelAndView getView(Object o, BindException errors)
-        {
-            return new JspView<>("/org/labkey/microarray/view/ExpressionMatrixQuery.jsp");
```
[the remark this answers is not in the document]
RosalineP: Done!

microarray/webapp/WEB-INF/microarray/microarrayContext.xml
RosalineP: Done!

microarray/src/org/labkey/microarray/query/MicroarrayUserSchema.java
labkey-jeckels: This method isn't used in module currently. Existing use cases are using getTable() which calls createTable() instead of going through this wrapper method. I think it's safe and good to delete it, especially since it calls the deprecated version of getTable() at the moment.

microarray/src/org/labkey/microarray/MicroarrayManager.java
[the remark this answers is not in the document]
labkey-jeckels: Similar to the getAnnotationTable() in the schema class, this method isn't used in the current code either. It's just a convenience wrapper over the standard getTable() method on the schema, so I think it's fair game to delete.

microarray/src/org/labkey/microarray/query/MicroarrayUserSchema.java
RosalineP: Cool, sounds like this is good to go then.

microarray/src/org/labkey/microarray/MicroarrayModule.java
RosalineP: Good catch, I've pruned out a couple more (most notably that MageML) that weren't doing anything. Everything else looks necessary.

.gitattributes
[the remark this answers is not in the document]
RosalineP: All deleted!

microarray/src/org/labkey/microarray/MicroarrayModule.java
```diff
@@ -50,38 +43,21 @@
...
     public static final AssayDataType THUMBNAIL_INPUT_TYPE =
             new AssayDataType("MicroarrayImageData", new FileType(".jpg"), "ThumbnailImage");
-    public static final AssayDataType FEATURES_INPUT_TYPE =
-            new AssayDataType("MicroarrayFeaturesData", new FileType("_feat.csv"), "Features");
-    public static final AssayDataType GRID_INPUT_TYPE =
-            new AssayDataType("MicroarrayGridData", new FileType("_grid.csv"), "Grid");
-
-    /** Collection of all of the non-MageML input types that are handled specially in the code */
-    public static final List<AssayDataType> RELATED_INPUT_TYPES =
-            Arrays.asList(QC_REPORT_INPUT_TYPE, THUMBNAIL_INPUT_TYPE, FEATURES_INPUT_TYPE, GRID_INPUT_TYPE);
```
labkey-jeckels: Shouldn't getVersion() return 19.21?

RosalineP: Ah, I misread this bit in the docs. Thank you, updated!

PR #55, opened 2019-08-19
6160: HTML JSP Warnings

RosalineP:


flow/src/org/labkey/flow/controllers/executescript/importAnalysisReviewSamples.jsp
```diff
@@ -35,6 +35,7 @@
 <%@ page import="java.util.List" %>
 <%@ page import="java.util.Map" %>
 <%@ page import="java.util.TreeMap" %>
+<%@ page import="org.labkey.api.util.HtmlString" %>
```
labkey-adam: Remove unused import?

luminex/src/org/labkey/luminex/view/guideSetConfirmDelete.jsp
```diff
@@ -60,7 +61,7 @@
             <br>
             Type: <% if(gs.isValueBased()) out.print("Value-based"); else out.print("Run-based"); %>
             <br><br>
-            Current Guide Set: <%=gs.getCurrent()%>
+            Current Guide Set: <%=HtmlString.of(gs.getCurrent().toString())%>
```
labkey-adam: I would just call h() on Booleans. It's simpler... plus it handles the case where gs.getCurrent() is null. (This code will NPE.)

labkey-adam: We could consider adding an h(Boolean) variant to JspBase, but I think the h(Object) variant is essentially the same thing.

RosalineP: Good point!

flow/src/org/labkey/flow/controllers/executescript/importAnalysisReviewSamples.jsp
RosalineP: Oops, thanks for the catch!

luminex/src/org/labkey/luminex/view/guideSetConfirmDelete.jsp
```diff
@@ -23,6 +23,7 @@
 <%@ page import="org.labkey.luminex.LuminexController" %>
 <%@ page import="java.util.List" %>
 <%@ page import="org.labkey.api.view.ActionURL" %>
+<%@ page import="org.labkey.api.util.HtmlString" %>
```
labkey-adam: Now this is an unused import :) But let's not worry about it...

PR #57, opened 2019-08-21
6160: HTML JSP Warnings 

RosalineP:


ms2/src/org/labkey/ms2/proteinSequence.jsp
```diff
@@ -38,4 +39,4 @@
         </tr>
     </table>
 <% } %>
-<big><tt><%= bean.protein.getFormattedSequence(bean.run) %></tt></big>
+<big><tt><%= h(bean.protein.getFormattedSequence(bean.run)) %></tt></big>
```
labkey-adam: Look at getFormattedSequence() -- it's producing HTML (notice startTag and endTag). Wrapping it with h() will not have the desired effect.

The method has only one call, so I suggest you update it to return an HtmlString, wrapping that StringBuffer in HtmlString.unsafe(). While you're at it, replace StringBuffer with StringBuilder... should equivalent, and there's no need for the concurrency guarantees that StringBuffer provides. (Let me know if you want me to explain.)

RosalineP: Ah, I didn't catch those HTML startTag and endTags... thanks for the suggestions, implemented.

PR #58, opened 2019-08-21
Fb jsp warnings 6160 daily b

ilikepies101:
This PR takes care of the warnings generated when the Daily B test suite is run that are not already covered in the BVT run (mostly strings / stringBuilders).

### DIFF
--- a/flow/src/org/labkey/flow/controllers/well/chooseGraph.jsp
+++ b/flow/src/org/labkey/flow/controllers/well/chooseGraph.jsp
@@ -116,7 +116,7 @@
         <% if (hasScripts)
         {
         %>
-        <tr><td>Analysis Script:</td><td><select name="<%=HtmlString.of(FlowParam.scriptId.toString())%>" onchange="this.form.submit()">
+        <tr><td>Analysis Script:</td><td><select name="<%=h(FlowParam.scriptId)%>" onchange="this.form.submit()">
             <labkey:options value="<%=form.getScriptId()%>" map="<%=scriptOptions%>"/>
         </select></td></tr>
         <%
--- a/flow/src/org/labkey/flow/controllers/well/chooseGraph.jsp
+++ b/flow/src/org/labkey/flow/controllers/well/chooseGraph.jsp
@@ -116,7 +116,7 @@
         <% if (hasScripts)
         {
         %>
-        <tr><td>Analysis Script:</td><td><select name="<%=h(FlowParam.scriptId)%>" onchange="this.form.submit()">
+        <tr><td>Analysis Script:</td><td><select name="<%=FlowParam.scriptId%>" onchange="this.form.submit()">
             <labkey:options value="<%=form.getScriptId()%>" map="<%=scriptOptions%>"/>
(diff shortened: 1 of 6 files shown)


flow/src/org/labkey/flow/controllers/well/chooseGraph.jsp
```diff
@@ -115,7 +116,7 @@
         <% if (hasScripts)
         {
         %>
-        <tr><td>Analysis Script:</td><td><select name="<%=FlowParam.scriptId%>" onchange="this.form.submit()">
+        <tr><td>Analysis Script:</td><td><select name="<%=h(FlowParam.scriptId)%>" onchange="this.form.submit()">
```
labkey-adam: Should FlowParam implement EnumHasHtmlString? Then h() would be unnecessary.

flow/src/org/labkey/flow/controllers/well/showWell.jsp
```diff
@@ -305,7 +305,7 @@ function showKeywords()
     for (var i=0 ; i<keywords.length ; i++)
     {
         var o = keywords[i];
-        o.label = o.index == 0 ? 'Keywords' : 'Parameter ' + _pad(o.index) + ' -- ' + o.name;
+        o.label = o.index == 0 ? h('Keywords') : h('Parameter ' + _pad(o.index) + ' -- ' + o.name);
```
labkey-adam: This is JavaScript code in the middle of a JSP! h() is not going to work there. I don't think you need any changes here.

flow/src/org/labkey/flow/controllers/well/showWell.jsp
```diff
@@ -348,10 +348,10 @@ Ext.onReady(function()
     }
 });
 
-var statistics = <%=jsonStats%>;
+var statistics = <%=unsafe(jsonStats.toString())%>;
 var treeData;
 var stats;
-var keywords = <%=jsonKeywords%>;
+var keywords = <%=unsafe(jsonKeywords.toString())%>;
```
labkey-adam: No change needed here... but we really should make unsafe() and h() accept CharSequence instead of String. That would eliminate the need for these toString() calls. Soon...

ms2/src/org/labkey/ms2/proteinSequence.jsp
```diff
@@ -38,4 +38,4 @@
         </tr>
     </table>
 <% } %>
-<big><tt><%= bean.protein.getFormattedSequence(bean.run) %></tt></big>
+<big><tt><%= h(bean.protein.getFormattedSequence(bean.run)) %></tt></big>
```
labkey-adam: Rosaline had this change as well. Look at getFormattedSequence()... it generates HTML (see the tag constants). She already changed that method to return HtmlString, so you can revert this change.

flow/src/org/labkey/flow/controllers/well/chooseGraph.jsp
ilikepies101: It should!

PR #60, opened 2019-08-29
Fb jsp finder

labkey-adam:


labkey-jeckels:
Did a cursory review and all looks fine to me. Since most of this is flow-related changes, I suggest that @labkey-kevink take a look too.

labkey-adam:
Agreed. FYI: All flow tests pass on this FB:

- https://teamcity.labkey.org/viewLog.html?buildId=806388
- https://teamcity.labkey.org/viewLog.html?buildId=806381

PR #62, opened 2019-09-04
6160: Final JSPs

RosalineP:


flow/src/org/labkey/flow/reports/statPicker.jsp
```diff
@@ -93,16 +96,9 @@
...
-    for (CompareType ct : new CompareType[] { CompareType.EQUAL, CompareType.NEQ_OR_NULL, CompareType.ISBLANK, CompareType.NONBLANK, CompareType.GT, CompareType.LT, CompareType.GTE, CompareType.LTE, CompareType.CONTAINS, CompareType.STARTS_WITH, CompareType.DOES_NOT_CONTAIN, CompareType.DOES_NOT_START_WITH, CompareType.IN })
-    {
-        ops.append(comma);
-        ops.append("[\"").append(ct.getPreferredUrlKey()).append("\", \"").append(ct.getDisplayValue()).append("\"]");
-        comma = ",\n";
-    }
-    ops.append("]");
+    var compareTypesToDisplay = Stream.of(CompareType.EQUAL, CompareType.NEQ_OR_NULL, CompareType.ISBLANK, CompareType.NONBLANK, CompareType.GT, CompareType.LT, CompareType.GTE, CompareType.LTE, CompareType.CONTAINS, CompareType.STARTS_WITH, CompareType.DOES_NOT_CONTAIN, CompareType.DOES_NOT_START_WITH, CompareType.IN )
+            .map(ct -> new String[] {ct.getPreferredUrlKey(), ct.getDisplayValue()});
+    JSONArray ops = toJsonArray(compareTypesToDisplay.collect(Collectors.toList()));
```
labkey-adam: Why not use toJsonArray(Stream) here?

(Even cooler would be to implement a JSONArrayCollector. See Collectors.of(). I'm not really serious... Collectors are a bit mind-bending to implement. See MultiValuedMapCollectors.java for an example.)

ms2/src/org/labkey/ms2/MS2Controller.java
```diff
@@ -606,39 +608,48 @@ private CurrentFilterBean(String[] headers, List<Pair<String, String>> sqlSummar
...
-    private StringBuilder renderViewSelect(boolean selectCurrent)
+    private Select.SelectBuilder renderViewSelect(boolean selectCurrent)
     {
         Map<String, String> m = getViewMap(true, getContainer().hasPermission(getUser(), ReadPermission.class));
 
-        StringBuilder viewSelect = new StringBuilder("<select id=\"views\" name=\"viewParams\" style=\"width:200\">");
+        Select.SelectBuilder select = new Select.SelectBuilder()
+                                        .id("views")
+                                        .name("viewParams")
+                                        .attributes(Collections.singletonMap("style", "width:200"));
```
labkey-adam: Obviously needs to change if you change SelectBuilder to have .style()

ms2/src/org/labkey/ms2/proteinCoverageMap.jsp
```diff
@@ -32,4 +32,4 @@ LABKEY.requiresScript("util.js");
     Ext.QuickTips.init();
 </script>
 
-<%=bean.protein.getCoverageMap(bean.run, bean.showRunUrl, bean.aaRowWidth) %>
+<%=unsafe(bean.protein.getCoverageMap(bean.run, bean.showRunUrl, bean.aaRowWidth).toString())%>
```
labkey-adam: Actually... maybe... getCoverageMap() should return HtmlString. That pushes unsafe() closer to the rendering code. There are a few other usages that will (right now) need to toString() that HtmlString to render it, but it gets us closer to where we want to be.

flow/src/org/labkey/flow/reports/statPicker.jsp
labkey-adam: Sign... I "accidentally" implemented JSONArray.collector(). I'll commit it to this branch...

ilikepies101: Haha nice! I also tried to use JsonArray(Stream) but it didn't work too well since the object passed in is `Stream<String[]>` and it's looking for a `Stream<Object>`. I cast the array as an object though (which looked a little funky), but it worked.

PR #71, opened 2019-09-17
6221: Luminex Cross-Plate Titrations

RosalineP:


RosalineP:
Oop, I broke some single-point tests.
Looking at those now.

luminex/src/org/labkey/luminex/LuminexExcelParser.java
```diff
@@ -242,6 +253,93 @@ private void parseFile() throws ExperimentException
...
+
+                // Check for single-file titrations. Does not check for dilution counts.
+                if (potentialTitrationCounts.get(desc) != null)
+                {
+                    // To Reviewer: It's possible to make the runtime better by removing the description from consideration in the main function's potentialTitration(Raw or Summary)Counts
+                    // once we recognize that it is a titration. However I found this makes debugging a bit harder / perhaps the code less readable because the description count is
+                    // no longer there/updated and accurate to the data.
+                    mini.put(analyteName, potentialTitrationCounts.get(desc));
+                    crossFilePT.put(desc, mini);
+                    if ((type.contentEquals("raw")) && ((potentialTitrationCounts.get(desc)) >= LuminexDataHandler.MINIMUM_TITRATION_RAW_COUNT))
```
cnathe: I don't think I have ever used the `contentEquals()` method for string comparison before. Was there a specific reason to use it here over just using `equals()` or `equalsIgnoreCase()`?

luminex/src/org/labkey/luminex/LuminexExcelParser.java
```diff
@@ -242,6 +253,93 @@ private void parseFile() throws ExperimentException
...
+                else
+                {
+                    crossFilePT.get(desc).put(analyteName, potentialTitrationCounts.get(desc));
+                }
+            }
+            else
+            {
+                HashMap<String, Integer> mini = new HashMap<>();
+
+                // Check for single-file titrations. Does not check for dilution counts.
```
cnathe: I would probably factor out this single-file titration check and call it above where you call `buildCrossFilePTs()`. Then this method doesn't have two functions, there would be one for building the cross file PTs and one for checking the single file PTs.

luminex/src/org/labkey/luminex/LuminexExcelParser.java
```diff
@@ -242,6 +253,93 @@ private void parseFile() throws ExperimentException
...
+                if (potentialTitrationCounts.get(desc) != null)
+                {
+                    // To Reviewer: It's possible to make the runtime better by removing the description from consideration in the main function's potentialTitration(Raw or Summary)Counts
+                    // once we recognize that it is a titration. However I found this makes debugging a bit harder / perhaps the code less readable because the description count is
+                    // no longer there/updated and accurate to the data.
```
cnathe: Thanks for the note. I agree that checking this each time even when the desc might already be in the _titrations set isn't ideal. But that runtime performance improvement would be so small that I think this is fine to do each time.

luminex/src/org/labkey/luminex/LuminexExcelParser.java
```diff
@@ -242,6 +253,93 @@ private void parseFile() throws ExperimentException
...
+            {
+                if (crossFilePT.get(desc).containsKey(analyteName))
+                {
+                    Integer count = crossFilePT.get(desc).get(analyteName);
+                    crossFilePT.get(desc).put(analyteName, count + potentialTitrationCounts.get(desc));
+                }
+                else
+                {
+                    crossFilePT.get(desc).put(analyteName, potentialTitrationCounts.get(desc));
+                }
```
cnathe: If we do the refactor suggested above about removing the analyte name part of the map, this can probably be simplified to just make sure the given desc is in the crossFilePT map and then add on the counts for that desc.

luminex/src/org/labkey/luminex/LuminexExcelParser.java
```diff
@@ -242,6 +253,93 @@ private void parseFile() throws ExperimentException
...
+            {
+                for (String analyte : crossFilePT.get(desc).keySet())
+                {
+                    boolean foundTitration = false;
+                    if ((crossFilePT.get(desc).get(analyte) >= minimum_titration_count) && (dilutionCounts.get(desc).size() >= 3))
+                    {
+                        _titrations.put(desc, crossFilePTs.get(desc));
+                        foundTitration = true;
+                    }
+                    // to reviewer: Realized I didn't ask this. Are cross-plate point controls something that is done? If not, we can remove this chunk.
```
cnathe: No. I don't think so. I believe only titrations would need to be considered for the cross plate checks, so I agree this can be removed.

luminex/src/org/labkey/luminex/LuminexExcelParser.java
```diff
@@ -242,6 +253,93 @@ private void parseFile() throws ExperimentException
...
+        int single_point_control_count = (type.contentEquals("raw")) ? LuminexDataHandler.SINGLE_POINT_CONTROL_RAW_COUNT : LuminexDataHandler.SINGLE_POINT_CONTROL_SUMMARY_COUNT;
+
+        for (String desc : crossFilePTs.keySet())
+        {
+            if (crossFilePT.containsKey(desc))
+            {
+                for (String analyte : crossFilePT.get(desc).keySet())
+                {
+                    boolean foundTitration = false;
+                    if ((crossFilePT.get(desc).get(analyte) >= minimum_titration_count) && (dilutionCounts.get(desc).size() >= 3))
```
cnathe: I believe in the cross plate titration check we only want to be considering the distinct dilution count for the given description values. So I think this if statement can remove the first clause and just check that `dilutionCounts.get(desc).size() >= 3`. Also, I think instead of hardcoding that value to 3, we should move this to a constant like the `MINIMUM_TITRATION_RAW_COUNT`

luminex/src/org/labkey/luminex/LuminexExcelParser.java
```diff
@@ -242,6 +253,93 @@ private void parseFile() throws ExperimentException
...
+    private void checkCrossFilePTs(Map<String, HashMap<String, Integer>> crossFilePT, Map<String, Titration> crossFilePTs, Map<String, HashSet<Double>> dilutionCounts, String type)
+    {
+        int minimum_titration_count = (type.contentEquals("raw")) ? LuminexDataHandler.MINIMUM_TITRATION_RAW_COUNT : LuminexDataHandler.MINIMUM_TITRATION_SUMMARY_COUNT;
+        int single_point_control_count = (type.contentEquals("raw")) ? LuminexDataHandler.SINGLE_POINT_CONTROL_RAW_COUNT : LuminexDataHandler.SINGLE_POINT_CONTROL_SUMMARY_COUNT;
+
+        for (String desc : crossFilePTs.keySet())
+        {
+            if (crossFilePT.containsKey(desc))
+            {
+                for (String analyte : crossFilePT.get(desc).keySet())
```
cnathe: With the refactor mentioned above to remove analyte from this map, this for loop will go away.

luminex/src/org/labkey/luminex/LuminexExcelParser.java
RosalineP: Done!

luminex/src/org/labkey/luminex/LuminexExcelParser.java
RosalineP: Gotcha!

luminex/src/org/labkey/luminex/LuminexExcelParser.java
```diff
@@ -86,12 +86,23 @@ private LuminexExcelParser(Domain excelRunDomain, Collection<File> dataFiles)
...
         if (_parsed) return;
         Set<String> analyteNames = new HashSet<>();
 
         for (File dataFile : _dataFiles)
         {
+            Map<String, Integer> potentialTitrationRawCounts = new CaseInsensitiveHashMap<>();
+            Map<String, Integer> potentialTitrationSummaryCounts = new CaseInsensitiveHashMap<>();
+            Map<String, Titration> potentialTitrations = new CaseInsensitiveHashMap<>();
+            boolean firstSheet = true;
             analyteNames.clear();
```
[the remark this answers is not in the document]
RosalineP: Done!

luminex/src/org/labkey/luminex/LuminexExcelParser.java
```diff
@@ -135,19 +146,27 @@ private void parseFile() throws ExperimentException
...
-
                     if (row <= sheet.getLastRowNum())
                     {
                         boolean hasMoreRows;
                         do
                         {
                             LuminexDataRow dataRow = createDataRow(analyte, sheet, colNames, row, dataFile);
 
-                            if (dataRow.getDescription() != null)
+                            if (firstSheet && dataRow.getDescription() != null)
```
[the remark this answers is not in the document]
RosalineP: Good point! Done.

luminex/src/org/labkey/luminex/LuminexExcelParser.java
RosalineP: Nope, I've changed them to equals() for consistency.

luminex/src/org/labkey/luminex/LuminexExcelParser.java
```diff
@@ -190,25 +208,13 @@ private void parseFile() throws ExperimentException
...
-                                    (potentialTitrationRawCounts.get(desc) != null && potentialTitrationRawCounts.get(desc) <= LuminexDataHandler.SINGLE_POINT_CONTROL_RAW_COUNT))
-                            {
-                                _singlePointControls.put(desc, new SinglePointControl(potentialTitrations.get(desc)));
-                            }
-                        }
-                    }
+                    firstSheet = false;
+
+                    crossFilePTRaw = buildCrossFilePTs(crossFilePTRaw, potentialTitrations, analyteName, potentialTitrationRawCounts, "raw");
+                    crossFilePTSummary = buildCrossFilePTs(crossFilePTSummary, potentialTitrations, analyteName, potentialTitrationSummaryCounts, "summary");
```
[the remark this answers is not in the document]
RosalineP: Done!

luminex/src/org/labkey/luminex/LuminexExcelParser.java
```diff
@@ -135,19 +146,27 @@ private void parseFile() throws ExperimentException
...
+                                if (dilutionCounts.containsKey(dataRow.getDescription()))
+                                {
+                                    dilutionCounts.get(dataRow.getDescription()).add(dataRow.getDilution());
+                                }
+                                else
+                                {
+                                    HashSet<Double> dilutionValue = new HashSet<>();
+                                    dilutionValue.add(dataRow.getDilution());
+                                    dilutionCounts.put(dataRow.getDescription(), dilutionValue);
+                                }
```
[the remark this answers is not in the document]
RosalineP: Oh good point, done!

luminex/src/org/labkey/luminex/LuminexExcelParser.java
RosalineP: Done

luminex/src/org/labkey/luminex/LuminexExcelParser.java
```diff
@@ -242,6 +253,93 @@ private void parseFile() throws ExperimentException
         _parsed = true;
     }
 
+    // Keeps track of potential titrations using the following structure. 'count_descriptionM_analyteN' indicates the number of instances where descriptionM and analyteN are found on the same row.
+    // crossFilePT: {description1: {analyte1: count_description1_analyte1, analtye2: count_description1_analyte2, ...},
+    //               description2: {analyte1: count_description2_analyte1, analyte2: count_description2_analyte2, ...} ... }
+    private Map<String, HashMap<String, Integer>> buildCrossFilePTs(Map<String, HashMap<String, Integer>> crossFilePT, Map<String, Titration> potentialTitrations, String analyteName, Map<String, Integer> potentialTitrationCounts, String type)
```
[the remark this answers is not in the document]
RosalineP: Ah, I'd wondered about that. I should have been more clear in describing what exactly what I thought got checked! 
Fix made.

luminex/src/org/labkey/luminex/LuminexExcelParser.java
RosalineP: I've added a MINIMUM_TITRATION_DILUTION_COUNT.
I've also taken out the first clause of the if statement -- and just to verify, that change implies that for cross-file potential titrations, we're only counting the dilution counts associated with the descriptions, and not the description counts across files.
If that is the case, then we don't need to be fully populating the crossFilePTs, just dilutionCounts.

luminex/src/org/labkey/luminex/LuminexExcelParser.java
RosalineP: Done!

luminex/src/org/labkey/luminex/LuminexExcelParser.java
cnathe: My take is that yes, cross plate titrations are found when a description exists in multiple files at 3 or more distinct dilutions, but single plate titrations are found when the description value exists more then 5 times in a single file. We should get buy in on this from Hannah (and from DHVI).

PR #72, opened 2019-09-18
Issue 37449: ensure the FlowProtocol exists

labkey-kevink:


flow/src/org/labkey/flow/controllers/WorkspaceData.java
```diff
@@ -200,7 +201,7 @@ else if (path.endsWith(".zip"))
 
                 _object = readWorkspace(file, path);
 
-                FlowProtocol protocol = FlowProtocol.getForContainer(container);
+                FlowProtocol protocol = FlowProtocol.ensureForContainer(user,  container);
```
labkey-susanh: Any problem with this validate method now having a side effect that writes to the database making this not useable for read-only scenarios?

labkey-tchad:
Looks good on TeamCity:
https://teamcity.labkey.org/viewLog.html?buildId=817863 (✔️)
https://teamcity.labkey.org/viewLog.html?buildId=817865 (✔️: One error is a known, intermittent problem)

PR #76, opened 2019-09-20
Fb migrate module deps

ilikepies101:


luminex/build.gradle
```diff
@@ -1,3 +1,6 @@
 import org.labkey.gradle.util.BuildUtils
 
 BuildUtils.addLabKeyDependency(project: project, config: "compile", depProjectPath: BuildUtils.getPlatformModuleProjectPath(project.gradle, "assay"), depProjectConfig: "apiCompile")
+BuildUtils.addLabKeyDependency(project: project, config: "modules", depProjectPath: BuildUtils.getPlatformModuleProjectPath(project.gradle, "experiment"), depProjectConfig: "published", depExtenstion: "module")
+BuildUtils.addLabKeyDependency(project: project, config: "modules", depProjectPath: BuildUtils.getPlatformModuleProjectPath(project.gradle, "list"), depProjectConfig: "published", depExtenstion: "module")
+BuildUtils.addLabKeyDependency(project: project, config: "modules", depProjectPath: BuildUtils.getPlatformModuleProjectPath(project.gradle, "pipeline"), depProjectConfig: "published", depExtenstion: "module")
```
labkey-klum: previous dependency was : experiment & assay, with your change it looks like it is now experiment, list, & pipeline?

ms2/build.gradle
```diff
@@ -3,6 +3,9 @@ import org.labkey.gradle.util.BuildUtils
 dependencies {
    external 'commons-httpclient:commons-httpclient:3.1'
    BuildUtils.addLabKeyDependency(project: project, config: "compile", depProjectPath: BuildUtils.getPlatformModuleProjectPath(project.gradle, "assay"), depProjectConfig: "apiCompile")
+   BuildUtils.addLabKeyDependency(project: project, config: "modules", depProjectPath: BuildUtils.getPlatformModuleProjectPath(project.gradle, "experiment"), depProjectConfig: "published", depExtenstion: "module")
+   BuildUtils.addLabKeyDependency(project: project, config: "modules", depProjectPath: BuildUtils.getPlatformModuleProjectPath(project.gradle, "pipeline"), depProjectConfig: "published", depExtenstion: "module")
+   BuildUtils.addLabKeyDependency(project: project, config: "modules", depProjectPath: BuildUtils.getPlatformModuleProjectPath(project.gradle, "experiment"), depProjectConfig: "published", depExtenstion: "module")
```
labkey-klum: experiment added twice?

ilikepies101: Good catch. Updated.

luminex/build.gradle
ilikepies101: Copy paste gone wrong - thanks!

viability/build.gradle
```diff
@@ -1,3 +1,8 @@
 import org.labkey.gradle.util.BuildUtils
 
-BuildUtils.addLabKeyDependency(project: project, config: "compile", depProjectPath: BuildUtils.getPlatformModuleProjectPath(project.gradle, "assay"), depProjectConfig: "apiCompile")
+dependencies{
```
labkey-susanh: These don't actually have to be within the dependencies closure, since the addLabKeyDependency method accesses this closure to do its business.  There might be a slight configuration performance implication, but if so it's probably slight, so if making other modifications you could remove these, but no big deal.

PR #96, opened 2019-11-07
Item 6493: Add createDomain to MassSpecFractionsDomainKind

labkey-martyp:
This was initially in AssayService but moving to domain kinds so assay creation can use DomainUtil validation which calls createDomain in the domain kinds.

### DIFF
--- a/ms2/src/org/labkey/ms2/metadata/MassSpecFractionsDomainKind.java
+++ b/ms2/src/org/labkey/ms2/metadata/MassSpecFractionsDomainKind.java
@@ -17,15 +17,21 @@ package org.labkey.ms2.metadata;
 
 import org.jetbrains.annotations.Nullable;
 import org.labkey.api.data.Container;
+import org.labkey.api.exp.DomainDescriptor;
 import org.labkey.api.exp.Lsid;
 import org.labkey.api.assay.AssayDomainKind;
+import org.labkey.api.exp.OntologyManager;
+import org.labkey.api.exp.TemplateInfo;
 import org.labkey.api.exp.property.Domain;
+import org.labkey.api.exp.property.PropertyService;
+import org.labkey.api.gwt.client.model.GWTDomain;
 import org.labkey.api.security.User;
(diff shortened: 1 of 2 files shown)


ms2/src/org/labkey/ms2/metadata/MassSpecFractionsDomainKind.java
```diff
@@ -116,4 +122,14 @@ public void appendNavTrail(NavTree root, Container c, User user)
...
     }
+
+    @Override
+    public Domain createDomain(GWTDomain domain, Map<String, Object> arguments, Container container, User user, @Nullable TemplateInfo templateInfo)
+    {
+        DomainDescriptor dd = OntologyManager.ensureDomainDescriptor(domain.getDomainURI(), domain.getName(), container);
+        dd = dd.edit().setDescription(domain.getDescription()).build();
+        OntologyManager.updateDomainDescriptor(dd);
+
+        return PropertyService.get().getDomain(container, dd.getDomainURI());
```
cnathe: Same comment from the PlateBasedAssaySampleSetDomainKind.java

Many of the other overrides in this java class use _assayDelegate, could this one? So I think it would be:
`return _assayDelegate.createDomain(domain, arguments, container, user, templateInfo);`

labkey-jeckels: I don't understand how createDomain() is working for assays in general. AbstractDomainKind's implementation returns null, and AssayBatchDomainKind, AssayRunDomainKind, etc don't override it.

cnathe: @labkey-martyp You can likely answer this better, but were the createDomain() pieces added to these PRs because of the error handling case? That is, you wanted to create the domain object so that you can use the standard domain validation code checks (and since this happens in a transaction they will get rolled back if errors exist)? I assume the AssayBatchDomainKind and AssayRunDomainKind didn't need it because the assay provider has a notion of createDefaultDomains().

PR #105, opened 2019-12-20
Remove GXT dependencies

labkey-klum:


labkey-klum:
Not for this particular change, although the change in platform (plate designer) required a change to the elispotAssayTest in testAutomation

PR #107, opened 2019-12-31
Issue 39299: Assay result row and sample count metrics

labkey-jeckels:


cnathe:
The GPAT assay would be the only one we care about for the Sample Manager case, but would there be any other assay providers outside of the platform and commonAssays ones that we care about for tracking for say Biologics?

labkey-kevink:
I'm not sure of the intent of the metric but my guess is that the result
row count for the flow data should be grouped per well instead of counting
all statistic value rows.  I can help after the holidays.

Kevin

labkey-jeckels:
@labkey-kevink The intent is to get better insight into adoption of assays. I don't think it's critical for the counts to match what you'd see in the schema browser, and I'd like for these queries to be as cheap as possible to run. I'll go ahead and merge this as-is, and resolve the tracking issue to you to consider followup changes on the flow side at your convenience

PR #128, opened 2020-02-20
Issue 39108: LuminexDataHandler throw more meaningful error message then NPE for missing dilution value when calculating obsConc

cnathe:


cnathe:
@labkey-jeckels Josh, I'd like your input on this one. The issue is that the user got an NPE stacktrace when uploading a Luminex assay run because of a null dilution value when the LuminexDataHandler tried to use that dilution to calculate obsConc. 

In this users case the empty dilution value was because of an error in their R transform script (which they have since fixed), but I was able to repro this by deleting the dilution value from one of our test data files ("plate 1_IgA-Biot (Standard2).xls" analyte ENV4 well D8). 

I don't think that having a null dilution value is actually valid for the Bioplex Luminex data format, so I think the "right" fix is to add a null check to LuminexDataHandler lines 1770 and 1774 (like the null check at line 1743 and 1754). However, doing that would have meant that this user would have gotten a successful data upload with their errant R transform script. 

This PR at least makes that error message more meaningful then an NPE, but it still throws an exception, which is not ideal. Thoughts?

labkey-jeckels (approved):
Looks good to me. Sounds like there's no legit use cases for having a null dilution value, and given the NPE we can be sure that we haven't had any quietly imported in the past.

PR #131, opened 2020-02-29
39821: Stash workspaces in session

labkey-adam:


flow/src/org/labkey/flow/controllers/editscript/UploadAnalysisForm.java
```diff
@@ -58,10 +60,26 @@ public void reset()
...
+        return token;
+    }
+
+    // On form POST, get the workspace from session using the token
+    public void setToken(String token)
+    {
+        this.token = token;
+        if (token != null)
+        {
+            _workspaceObject = (Workspace)SessionHelper.getStashedAttribute(getRequest(), token);
```
labkey-matthewb, line 79: If token lookup fails the validate() method should generate a meaningful error message.

labkey-matthewb (changes requested):

labkey-matthewb (approved):

PR #139, opened 2020-03-19
Update dependencies for labkey-client-api to use proper group and version number

labkey-susanh:


labkey-klum (approved):

labkey-susanh:
targeting 20.3-SNAPSHOT instead

PR #166, opened 2020-05-05
add validation to avoid IllegalPathException

labkey-matthewb:
#### Rationale
avoid uncaught exception


labkey-jeckels (approved):
Thanks for the fix - a crawler-based test failure was assigned to me today that this should address

PR #169, opened 2020-05-19
Issue 40151: Show modifications shows HTML code instead of a table of mods

labkey-jeckels:
#### Changes
Avoid double-encoding and use DOM builders

### DIFF
--- a/ms2/src/org/labkey/ms2/MS2Controller.java
+++ b/ms2/src/org/labkey/ms2/MS2Controller.java
@@ -190,6 +190,13 @@ import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.TreeSet;
+import java.util.stream.Stream;
+
+import static org.labkey.api.util.DOM.STRONG;
+import static org.labkey.api.util.DOM.TABLE;
+import static org.labkey.api.util.DOM.TD;
+import static org.labkey.api.util.DOM.TR;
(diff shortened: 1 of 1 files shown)


labkey-matthewb (approved):
It's a bit more compact if you do import static org.labkey.api.util.DOM.*

Also, I'd encourage you to try to build this all as one statement starting with DOM.TABLE(...) with no ArrayList<> rows.

* you can pass null as a noop, if there is a conditional element (e.g. an optional row)
    ..., cond ? null : TR(TD()), ...
* you can return fragments from a method (list of elements with no wrapping element)
   Renderable appendMods() { ... return DOM.createHtmlFragment(...)}
* These methods can take streams as arguments so.
   createHtmlFragment( mods.entrySet().stream().map(entry -> DOM.TR(...)) )

Approved in case you're not in the mood.

labkey-matthewb:
ModulesView.renderView() is an extreme example

labkey-jeckels:
Thanks Matt, I was hoping you'd have some good tips.  If you don't mind a quick re-review, I think I've updated as suggested.

It's tricky to discover what's a legal argument since most methods take Object... I'll add some JavaDoc that tries to capture these tips.

labkey-matthewb:
I know, I had to pick between varargs... or overloads, can't really do both...

labkey-matthewb (approved):

PR #191, opened 2020-06-18
Skip mock mascot test if devtools module is missing

labkey-tchad:
#### Rationale
DeveloperTools module might not be present on production servers. This change allows us to run more tests on remote servers, which tend to be running in production mode.

#### Changes
* Skip mock mascot test in production mode if the `devtools` modules isn't present.

### DIFF
--- a/ms2/src/org/labkey/ms2/pipeline/mascot/MascotClientImpl.java
+++ b/ms2/src/org/labkey/ms2/pipeline/mascot/MascotClientImpl.java
@@ -27,12 +27,14 @@ import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClients;
 import org.apache.log4j.Logger;
 import org.junit.Assert;
+import org.junit.Assume;
 import org.junit.Test;
 import org.labkey.api.module.ModuleLoader;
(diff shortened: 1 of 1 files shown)


labkey-susanh (approved):

ms2/src/org/labkey/ms2/pipeline/mascot/MascotClientImpl.java
```diff
@@ -1523,6 +1525,11 @@ private InputStream getRequestResultStream (Properties parameters)
         @Test
         public void testMockMascotServer() throws IOException
         {
+            if (!AppProps.getInstance().isDevMode())
+            {
+                Assume.assumeNotNull("Mock mascot server requires DeveloperTools module. Skipping test in production mode.",
```
labkey-adam, line 1530: Wouldn't it be better to just check for module presence and skip the production mode check? It's possible to have a devmode build that lacks this module.

labkey-tchad: I've been following this pattern so that we don't start silently skipping tests when we shouldn't. Like, if that module's name changes without this getting updated.

PR #194, opened 2020-06-23
update jQuery reference in credits file

labkey-ankurj:
#### Rationale
Updates jQuery in credits file

#### Related Pull Requests
https://github.com/LabKey/platform/pull/1334
https://github.com/LabKey/targetedms/pull/213


labkey-jeckels (changes requested):
I think we can just remove this line, and have a single reference to it in the module where it's actually checked int

labkey-jeckels (approved):

PR #223, opened 2020-08-04
NAb table fixes to use QueryForeignKey for RunData and DilutionData related columns

cnathe:
#### Rationale
A couple of the FKs in the NAb assay schema tables were showing in the schema browser as "null.NAbSpecimen.RowId" and were also not allowing for all of the available calculated columns to be pulled into a custom view via that lookup. This was preventing a client scenario where they wanted to get the NAb assay DilutionData for all rows that were copied to a given study folder, but that "copied to study X" column could not be pulled into the DilutionData grid view.

Note: putting this into develop for now, TBD if this needs to be backported to another release.

#### Changes
* NabDilutionDataTable update to use QueryForeignKey for the RunData column lookup to the Data (i.e. NAbSpecimen) table
* NabWellDataTable update to use QueryForeignKey for the DilutionData and RunData column lookups

### DIFF
--- a/nab/src/org/labkey/nab/query/NabDilutionDataTable.java
+++ b/nab/src/org/labkey/nab/query/NabDilutionDataTable.java
@@ -19,6 +19,7 @@ import org.labkey.api.assay.dilution.DilutionManager;
 import org.labkey.api.data.ColumnInfo;
 import org.labkey.api.data.ContainerFilter;
 import org.labkey.api.exp.api.ExpProtocol;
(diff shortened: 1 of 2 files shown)


labkey-jeckels (approved):
This is targeting develop - is that correct?

cnathe:
> This is targeting develop - is that correct?

Yes, there is a workaround that we are seeing if the client is OK with in previous versions. So I thought I'd get this checked in and we can backport it if needed.

PR #224, opened 2020-08-07
Eliminate JSP print(String) warnings from community module test suites

labkey-adam:
#### Rationale
We want to eliminate rendering of raw Strings from JSPs. See platform PR for more details.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1470


flow/src/org/labkey/flow/controllers/editscript/editCompensationCalculation.jsp
```diff
@@ -76,7 +75,7 @@ var SS = []; // SUBSETS
             if (!hashToIndexMap.containsKey(subsetHash))
             {
                 hashToIndexMap.put(subsetHash, index);
-                index = index.intValue() + 1;
+                index = (int)index + 1;
```
labkey-jeckels, line 78: This cast looks redundant

ms2/src/org/labkey/ms2/showPeptide.jsp
```diff
@@ -66,7 +65,9 @@
 <%
         if (fraction.wasloadedFromGzFile())
         {
-            out.println(" " + MS2GZFileRenderer.getFileNameInGZFile(fraction.getFileName(), p.getScan(), p.getCharge(), "dta") + "<br>");
+            out.print(unsafe(" "));
```
labkey-jeckels, line 68: Would it make sense to introduce constants in HtmlString for normal spaces and <br> tags, like we have for non-breaking spaces? Seems like there are few places that are now doing unsafe() on them

labkey-jeckels (approved):
A couple of suggestions but doesn't require additional review

flow/src/org/labkey/flow/controllers/editscript/editCompensationCalculation.jsp
labkey-adam: I noticed that as well... already fixed it in my Daily change list. IntelliJ made the Integer → int suggestion, odd that it would add the cast.

ms2/src/org/labkey/ms2/showPeptide.jsp
labkey-adam: Yep, good idea. I don't want to overlap too much with DOM, but space and the most common tags would make sense.

PR #226, opened 2020-08-24
Issue 41224: support duplicate sample names in FlowJo workspace

labkey-kevink:
#### [Issue 41224](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41224): support duplicate sample names in FlowJo workspace

The flow import process supports duplicate samples for external analysis archives but not for FlowJo workspaces.  This change consolidates sample handling in the workspace data model to ensure unique sample IDs but allow duplicate sample labels across all format types.

The FCS file resolve step in the import wizard switched from using a simple Map of sample label to the resolved FCS file to a SampleIdMap to guarantee unique sample ID values.  The import wizard supports selecting FCS files from the workspace for import that aren't resolved in two cases: (1) when not resolving against any FCS files, (2) importing a directory of FCS files at the same time as the workspace is imported.  In the latter case, the newly imported FCS files will be matched against the samples in the workspace by keyword values.  A marker FlowFCSFile.UNMATCHED value is used to represent the selected but unresolved samples that will be imported.

#### [Issue 41225](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41225): flow: import failure for duplicate aliased statistics

It is possible to have two boolean populations defined that result in the same alias. When importing the workspace, the boolean population's count statistic will have an alias created that represents the boolean expression. If two boolean populations have the same boolean expression aliases, they will all become aliases of each other. For example, the count statistic of boolean population "Foo" that represents A or B will have an alias created "(A|B):Count" for "Foo:Count". If a second population "Bar" also represents A or B, it will have the same alias "(A|B):Count" and will become aliases of "Foo:Count" as well.

When importing the statistics, we insert the stat value using the preferred statistic name ("Foo:Count" in our example) and enforce a single statistic definition for each flow well.  This change detects duplicate statistics and will print a warning to the job log file if the statistic values are identical or throw an error if they are not.

#### Related Pull Requests
* https://github.com/LabKey/testAutomation/pull/441

### DIFF
--- a/flow/enginesrc/org/labkey/flow/analysis/model/BaseWorkspace.java
+++ b/flow/enginesrc/org/labkey/flow/analysis/model/BaseWorkspace.java
@@ -99,7 +99,7 @@ public abstract class BaseWorkspace<S extends ISampleInfo> implements IWorkspace
     @Override
     public List<S> getSamples()
     {
-        return new ArrayList<>(_sampleInfos.values());
+        return List.copyOf(_sampleInfos.values());
     }
 
     /**
--- a/flow/enginesrc/org/labkey/flow/analysis/model/BaseWorkspace.java
+++ b/flow/enginesrc/org/labkey/flow/analysis/model/BaseWorkspace.java
@@ -15,30 +15,37 @@
  */
 package org.labkey.flow.analysis.model;
 
+import org.jetbrains.annotations.NotNull;
 import org.labkey.api.collections.CaseInsensitiveMapWrapper;
 import org.labkey.api.collections.CaseInsensitiveTreeSet;
 import org.labkey.flow.persist.AttributeSet;
 
 import java.io.Serializable;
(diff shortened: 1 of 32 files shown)


flow/enginesrc/org/labkey/flow/analysis/model/BaseWorkspace.java
```diff
@@ -84,4 +91,92 @@ public AttributeSet getSampleAnalysisResults(ISampleInfo sample)
...
 
+    public int getSampleCount()
+    {
+        return _sampleInfos.size();
+    }
+
+    @Override
+    public List<S> getSamples()
+    {
+        return new ArrayList<>(_sampleInfos.values());
```
labkey-matthewb: use unmodifieableList()?

labkey-matthewb (approved):

PR #231, opened 2020-09-15
Adjust all JSPs to extend JspContext

labkey-adam:
#### Rationale
All JSPs need to extend `JspContext`.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1564

### DIFF
--- a/ms2/src/org/labkey/ms2/compare/compareSearchEngineProteinOptions.jsp
+++ b/ms2/src/org/labkey/ms2/compare/compareSearchEngineProteinOptions.jsp
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 %>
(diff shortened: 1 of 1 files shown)


labkey-susanh (approved):
I'm guessing there will be a docs update somewhere so folks know of this extension need (not that we're writing new JSPs now, of course, but still).

labkey-adam:
It'll be pretty obvious when they attempt to compile their JSP for the first time :)

Seriously, it's very rare that a JSP doesn't extend JspBase... or at least JspContext. But I'll see if we have a JSP writing docs page somewhere.

labkey-adam:
Added some fodder (and fixed up some old code) in this topic: https://www.labkey.org/Documentation/wiki-page.view?name=views

PR #236, opened 2020-09-26
20.7 fb elisa 7669

labkey-klum:
#### Rationale
The ELISA assay is being enhanced to support a new high throughput data format which will introduce some new concepts:
- multiple plates per run (similar to the NAb high throughput variants)
- multiple values per well, through the use of analytes that could result in up to 4 values per well that need to be tracked separately
- multiple standard curves per plate, similar to the sample wells there could be multiple analytes for each plate well

Additionally, some new statistics will be added as detailed in this [spec](https://docs.google.com/document/d/1euUCUWOD-q5KESZsfiWLrTM4n6yaRuyHF2uCT2EXxWc/edit#). Run, sample, and result domains were expanded to include new data to be captured as well as to hold some of the new calculations. The ELISA data handler was refactored to handle the legacy and new behavior, this ensures that previous assays and new assays that don't use the new format will continue to work as before.

An experimental flag was put in place to hide the new features until a follow on story to update the run details report is complete.

#### Related Pull Requests
https://github.com/LabKey/platform/pull/1598

#### Changes
- Add assay provider option to collect both sample and results data from a combined file (assay design selection)
- Add run level option to specify the curve fit : 4 & 5 param, linear
- Refactor the data handler so it can process both legacy data input as well as the new high throughput format
- Add a 16 x 24 plate template which matches the NVITAL plate design
- Since linear curve fit is now a standard StatsService option, we need to constrain the NAb run option to not offer the new curve fit (NAbProviderSchema)
- Additionally for the ELISA provider schema, don't offer the polynomial fit option
- Push new features under an experimental flag

### DIFF
--- a/elisa/src/org/labkey/elisa/ElisaAssayProvider.java
+++ b/elisa/src/org/labkey/elisa/ElisaAssayProvider.java
@@ -308,7 +308,8 @@ public class ElisaAssayProvider extends AbstractPlateBasedAssayProvider
     {
         if (form instanceof ElisaRunUploadForm)
         {
-            return new JspView<>("/org/labkey/assay/view/tsvDataDescription.jsp", form);
+            if (((ElisaRunUploadForm)form).getSampleMetadataInputFormat() == SampleMetadataInputFormat.COMBINED)
+                return new JspView<>("/org/labkey/assay/view/tsvDataDescription.jsp", form);
         }
         return new HtmlView(HtmlString.of("The ELISA data files must be in the BioTek Microplate Reader Excel file format (.xls or .xlsx extension)."));
     }
--- a/elisa/src/org/labkey/elisa/ElisaAssayProvider.java
+++ b/elisa/src/org/labkey/elisa/ElisaAssayProvider.java
@@ -94,12 +94,12 @@ public class ElisaAssayProvider extends AbstractPlateBasedAssayProvider
 
     // results properties
     public static final String ABSORBANCE_PROPERTY = "Absorption";
+    public static final String STANDARD_CONCENTRATION_PROPERTY = "Std_Concentration";
     public static final String CONCENTRATION_PROPERTY = "Concentration";
     public static final String WELL_LOCATION_PROPERTY = "WellLocation";
     public static final String WELLGROUP_PROPERTY = "WellgroupLocation";
     public static final String SPOT_PROPERTY = "Spot";
     public static final String DILUTION_PROPERTY = "Dilution";
-    public static final String CV_PROPERTY = "CV";
     public static final String EXCLUDED_PROPERTY = "Excluded";
     public static final String PERCENT_RECOVERY = "Recovery_Percent";
     public static final String PERCENT_RECOVERY_MEAN = "Recovery_Mean";
@@ -114,12 +114,7 @@ public class ElisaAssayProvider extends AbstractPlateBasedAssayProvider
 
     static
     {
-        EXTRA_SAMPLE_PROPS = Set.of(
-                AUC_PROPERTY,
-                MEAN_ABSORPTION_PROPERTY,
-                CV_ABSORPTION_PROPERTY,
-                MEAN_CONCENTRATION_PROPERTY,
-                CV_CONCENTRATION_PROPERTY);
+        EXTRA_SAMPLE_PROPS = Set.of(AUC_PROPERTY);
     }
 
     enum PlateReaderType
@@ -237,11 +232,7 @@ public class ElisaAssayProvider extends AbstractPlateBasedAssayProvider
         addProperty(domain, DATE_PROPERTY_NAME, DATE_PROPERTY_CAPTION, PropertyType.DATE_TIME);
 
         // extra computed properties at the sample scope
-        addProperty(domain, AUC_PROPERTY, AUC_PROPERTY, PropertyType.DOUBLE, "Area Under the Curve");
-        addProperty(domain, MEAN_ABSORPTION_PROPERTY, "Absorption Mean", PropertyType.DOUBLE, "Mean Absorption across sample replicates");
-        addProperty(domain, CV_ABSORPTION_PROPERTY, "Absorption CV", PropertyType.DOUBLE, "Absorption CV across sample replicates");
-        addProperty(domain, MEAN_CONCENTRATION_PROPERTY, "Concentration Mean", PropertyType.DOUBLE, "Mean Concentration across sample replicates");
-        addProperty(domain, CV_CONCENTRATION_PROPERTY, "Concentration CV", PropertyType.DOUBLE, "Concentration CV across sample replicates");
+        addPropertyWithFormat(domain, AUC_PROPERTY, AUC_PROPERTY, PropertyType.DOUBLE, "Area Under the Curve", "0.0000");
 
         return result;
     }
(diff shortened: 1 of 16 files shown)


elisa/src/org/labkey/elisa/ElisaAssayProvider.java
```diff
@@ -79,25 +91,49 @@
...
+    public static final String MEAN_CONCENTRATION_PROPERTY = "Concentration_Mean";
+    public static final String CV_CONCENTRATION_PROPERTY = "Concentration_CV";
+    public static final String WELL_LOCATION_PROPERTY = "WellLocation";
+    public static final String WELLGROUP_PROPERTY = "WellgroupLocation";
+    public static final String SPOT_PROPERTY = "Spot";
+    public static final String DILUTION_PROPERTY = "Dilution";
+    public static final String EXCLUDED_PROPERTY = "Excluded";
+    public static final String PERCENT_RECOVERY = "Recovery_Percent";
+    public static final String PERCENT_RECOVERY_MEAN = "Recovery_Mean";
+    public static final String PLATE_PROPERTY = "PlateName";
```
cnathe, line 116: Was it intentional that some of the properties have underscores in them and others are just camelcase? Just curious if there is meaning to the split.

elisa/src/org/labkey/elisa/ElisaAssayProvider.java
```diff
@@ -183,11 +219,21 @@ public String getName()
         Pair<Domain, Map<DomainProperty, Object>> result = super.createRunDomain(c, user);
         Domain domain = result.getKey();
 
-        DomainProperty fitProp = addProperty(domain, CORRELATION_COEFFICIENT_PROPERTY_NAME, CORRELATION_COEFFICIENT_CAPTION, PropertyType.DOUBLE, "Coefficient of Determination of the calibration curve.");
+        DomainProperty fitProp = addProperty(domain, CORRELATION_COEFFICIENT_PROPERTY, "Coefficient of Determination", PropertyType.DOUBLE, "Coefficient of Determination of the calibration curve.");
         fitProp.setFormat("0.000");
         fitProp.setShownInInsertView(false);
+        addProperty(domain, CONCENTRATION_UNITS_PROPERTY, "Concentration Units", PropertyType.STRING, "Units eg. (ug/ml)");
```
cnathe, line 225: should this new property be behind the experimental feature flag?

elisa/src/org/labkey/elisa/ElisaAssayProvider.java
```diff
@@ -202,17 +248,21 @@ public String getName()
...
 
-        addProperty(domain, SPECIMENID_PROPERTY_NAME, SPECIMENID_PROPERTY_CAPTION, PropertyType.STRING);
+        DomainProperty specimenProperty = addProperty(domain, SPECIMENID_PROPERTY_NAME, SPECIMENID_PROPERTY_CAPTION, PropertyType.STRING);
+        specimenProperty.setImportAliasSet(Set.of("Sample"));
         addProperty(domain, PARTICIPANTID_PROPERTY_NAME, PARTICIPANTID_PROPERTY_CAPTION, PropertyType.STRING);
         addProperty(domain, VISITID_PROPERTY_NAME, VISITID_PROPERTY_CAPTION, PropertyType.DOUBLE);
         addProperty(domain, DATE_PROPERTY_NAME, DATE_PROPERTY_CAPTION, PropertyType.DATE_TIME);
 
+        // extra computed properties at the sample scope
+        addPropertyWithFormat(domain, AUC_PROPERTY, AUC_PROPERTY, PropertyType.DOUBLE, "Area Under the Curve", "0.0000");
```
cnathe, line 258: should this new property be behind the experimental feature flag?

elisa/src/org/labkey/elisa/ElisaAssayProvider.java
```diff
@@ -221,22 +271,53 @@ public String getName()
...
+        addProperty(dataDomain, WELLGROUP_PROPERTY, "Well Group", PropertyType.STRING, "Replicate Well Group");
 
-        DomainProperty absProp = addProperty(dataDomain, ABSORBANCE_PROPERTY_NAME,  ABSORBANCE_PROPERTY_CAPTION, PropertyType.DOUBLE, "Well group value measuring the absorption.");
-        absProp.setFormat("0.000");
-        DomainProperty concProp = addProperty(dataDomain, CONCENTRATION_PROPERTY_NAME,  CONCENTRATION_PROPERTY_CAPTION, PropertyType.DOUBLE, "Well group value measuring the concentration.");
-        concProp.setFormat("0.000");
+        addPropertyWithFormat(dataDomain, ABSORBANCE_PROPERTY,  "Absorption", PropertyType.DOUBLE, "Raw signal value", "0.000");
+        DomainProperty concProp = addPropertyWithFormat(dataDomain, CONCENTRATION_PROPERTY,  "Concentration", PropertyType.DOUBLE, "Calculated concentration", "0.000");
         concProp.setDefaultValueTypeEnum(DefaultValueType.LAST_ENTERED);
+        addPropertyWithFormat(dataDomain, STANDARD_CONCENTRATION_PROPERTY,  "Std Concentration", PropertyType.DOUBLE, "Standard control concentration", "0.000");
```
cnathe, line 280: should any of the new properties in this section be behind the experimental feature flag?

elisa/src/org/labkey/elisa/ElisaDataHandler.java
```diff
@@ -138,91 +147,133 @@ protected boolean shouldAddInputMaterials()
...
                                 }
 
-                                if (conc != -1)
-                                    row.put(ElisaAssayProvider.CONCENTRATION_PROPERTY_NAME, conc);
+                                if (!concentrations.isEmpty())
+                                {
+                                    // compute stats on the replicate values
+                                    MathStat concStat = StatsService.get().getStats(concentrations);
+                                    MathStat recovStat = StatsService.get().getStats(recoveries);
+                                    Double concCV = concStat.getStdDev() / concStat.getMean();
```
cnathe, line 200: Do we need the "divide by zero" check here for the `concCV`?

elisa/src/org/labkey/elisa/ElisaDataHandler.java
```diff
@@ -138,91 +147,133 @@ protected boolean shouldAddInputMaterials()
...
+                                }
 
-                                    // compute the concentration
-                                    double concentration = (well.getValue() - regression.getIntercept()) / regression.getSlope();
-                                    row.put(ElisaAssayProvider.CONCENTRATION_PROPERTY_NAME, concentration);
+                                if (!concentrations.isEmpty())
+                                {
+                                    // compute concentration mean and CV values
+                                    MathStat stat = StatsService.get().getStats(concentrations);
+                                    Double concCV = stat.getStdDev() / stat.getMean();
```
cnathe, line 246: Same question as above, do we care to check for "divide by zero"?

elisa/src/org/labkey/elisa/ElisaImportHelper.java
```diff
@@ -0,0 +1,54 @@
...
+    public static String PLACEHOLDER_PLATE_NAME = "PLACEHOLDER_PLATE";
+
+    /**
+     * Gets the set of plate names in this import
+     */
+    Set<String> getPlates();
+
+    /**
+     * Gets a map of analyte number to plate populated with raw signal values. Multi-plex plates will
+     * record multiple values per well (for each analyte). We will model this a multiple plates, one for
```
cnathe:
```suggestion
     * record multiple values per well (for each analyte). We will model this as multiple plates, one for
```

elisa/src/org/labkey/elisa/ElisaSampleFilePropertyHelper.java
```diff
@@ -0,0 +1,119 @@
...
+            boolean hasSpotColumn = Arrays.stream(loader.getColumns()).anyMatch(c -> c.getColumnName().equalsIgnoreCase(spotColumnName));
+
+            if (!hasPlateColumn)
+                throw new ExperimentException("Sample metadata file does not contain required column \"" + plateColumnName + "\".");
+            if (!hasSampleColumn)
+                throw new ExperimentException("Sample metadata file does not contain required column \"" + sampleColumnName + "\".");
+            if (!hasWellLocationColumn)
+                throw new ExperimentException("Sample metadata file does not contain required column \"" + wellLocationColumnName + "\".");
+            if (!hasSpotColumn)
+                throw new ExperimentException("Sample metadata file does not contain required column \"" + spotColumnName + "\".");
```
cnathe: This code to stream the columns to find a given column name and then throw if not found is replicated here and in `HighThroughputImportHelper`. Is there anywhere that would make sense to factor this out to shared code?

elisa/src/org/labkey/elisa/ElisaSampleFilePropertyHelper.java
```diff
@@ -0,0 +1,119 @@
...
+                if (wellLocation != null && plateName != null)
+                {
+                    String sampleWellGroup = getSampleWellGroupFromLocation(plateName, spot, wellLocation);
+                    if (sampleWellGroup != null)
+                    {
+                        Map<DomainProperty, String> sampleProperties = allProperties.computeIfAbsent(sampleWellGroup, k -> new HashMap<>());
+                        for (DomainProperty property : _domainProperties)
+                        {
+                            Object value = getValue(row, property);
+                            sampleProperties.put(property, value != null ? value.toString() : null);
```
cnathe, line 72: we are putting prop/values into `sampleProperties` here but then below we set `_sampleProperties = allProperties`, just checking to make sure this wasn't supposed to be a put into `allProperties`.

elisa/src/org/labkey/elisa/HighThroughputImportHelper.java
```diff
@@ -0,0 +1,304 @@
...
+                double[][] data = _dataMap.computeIfAbsent(spot, s -> new double[_plateTemplate.getRows()][_plateTemplate.getColumns()]);
+                data[position.getRow()][position.getColumn()-1] = signal.doubleValue();
+            }
+        }
+
+        public void setExtraProperties(Map<String, Object> row, PositionImpl position, Integer spot, List<? extends DomainProperty> resultsDomain)
+        {
+            Map<String, Object> extraProperties = new HashMap<>();
+            // need to adjust the column value to be 0 based to match the template locations
+            position.setColumn(position.getColumn()-1);
```
cnathe, line 225: this `position.getColumn()-1` code shows up a few times, is it worth adding a helper to the PositionImpl? something like position.getZeroBasedColumn() that just returns column-1?

elisa/src/org/labkey/elisa/HighThroughputImportHelper.java
```diff
@@ -0,0 +1,304 @@
...
+                {
+                    Object value = getValue(row, prop);
+                    if (value != null)
+                        extraProperties.put(prop.getName(), value);
+                }
+            }
+            _extraWellData.put(getWellAnalyteKey(position, spot), extraProperties);
+        }
+
+        private Object getValue(Map<String, Object> row, DomainProperty property)
```
cnathe, line 252: Do we have any sort of centralized place that does this "get value for a property name or import aliases"? It seems like something we'd possibly be doing all over the place when we have a row Map and a DomainProperty.

elisa/src/org/labkey/elisa/actions/ElisaUploadWizardAction.java
```diff
@@ -111,24 +112,34 @@ public boolean executeStep(ElisaRunUploadForm form, BindException errors) throws
...
 
     @Override
     protected void addRunActionButtons(ElisaRunUploadForm form, InsertView insertView, ButtonBar bbar)
     {
-        addNextButton(bbar);
+        // don't render the concentrations view
+        if (form.getSampleMetadataInputFormat() == SampleMetadataInputFormat.MANUAL)
+            addNextButton(bbar);
+        else
+            addFinishButtons(form, insertView, bbar);
```
cnathe, line 142: Likely not part of this PR changes, but I noticed in the UI that the final step of the ELISA assay wizard for the "Concentrations for Standard Wells", the button bar doesn't have the btn-primary styling for the "Save and Finish", etc. buttons.

cnathe (approved):
A few questions and comments, but nothing that will likely need re-review.

cnathe:
FYI, I just locally tried to upload one of the existing ELISA sampledata files, and I get the following exception on save and finish (with and without the experimental feature enabled):

```
java.lang.IllegalArgumentException: curveParameters must be an instance of LinearParameters
       at org.labkey.core.statistics.LinearCurveFit.solveForX(LinearCurveFit.java:102)
       at org.labkey.elisa.AbstractElisaImportHelper.createWellRow(AbstractElisaImportHelper.java:66)
       at org.labkey.elisa.ManualImportHelper.createWellRow(ManualImportHelper.java:91)
       at org.labkey.elisa.ElisaDataHandler.getValidationDataMap(ElisaDataHandler.java:163)
       at org.labkey.api.assay.AbstractAssayTsvDataHandler.importFile(AbstractAssayTsvDataHandler.java:124)
```

cnathe:
> FYI, I just locally tried to upload one of the existing ELISA sampledata files, and I get the following exception on save and finish (with and without the experimental feature enabled):
> 
> ```
> java.lang.IllegalArgumentException: curveParameters must be an instance of LinearParameters
>        at org.labkey.core.statistics.LinearCurveFit.solveForX(LinearCurveFit.java:102)
>        at org.labkey.elisa.AbstractElisaImportHelper.createWellRow(AbstractElisaImportHelper.java:66)
>        at org.labkey.elisa.ManualImportHelper.createWellRow(ManualImportHelper.java:91)
>        at org.labkey.elisa.ElisaDataHandler.getValidationDataMap(ElisaDataHandler.java:163)
>        at org.labkey.api.assay.AbstractAssayTsvDataHandler.importFile(AbstractAssayTsvDataHandler.java:124)
> ```

Update: when I re-imported the run and gave it more meaningful concentration values (i.e. 100 down to 0) instead of just setting all concentration values to 1, then the run uploaded successfully. So likely this means just a check is needed somewhere in the code when the concentration values are not descending (or something along those lines).

elisa/src/org/labkey/elisa/ElisaAssayProvider.java
labkey-klum: no, was mostly arbitrary

elisa/src/org/labkey/elisa/ElisaAssayProvider.java
labkey-klum: I was thinking this was general purpose and would apply to both the new and old assays

elisa/src/org/labkey/elisa/ElisaDataHandler.java
labkey-klum: I'll double check but it shouldn't throw but return something like Double.POSITIVE_INFINITY, which should be fine to put in the DB.

labkey-klum:
It probably means the standard concentrations couldn't be fit to a curve, I'll add a check for that.

PR #239, opened 2020-09-30
ELISA assay run details - fix issue to restore display of run details information panel above plot

cnathe:
#### Rationale
As we are about to update the ELISA assay run details page, I thought it would be good to fix the current issue that was preventing the run details information panel from showing on the current run details page. This way we know where we are actually starting from before making changes.

#### Changes
* simplify the data loading logic for the Ext4 store
* fix JS error with Ext.util.Format.number instead of Ext4.util.Format.number

### DIFF
--- a/elisa/webapp/elisa/runDetailsPanel.js
+++ b/elisa/webapp/elisa/runDetailsPanel.js
@@ -22,15 +22,29 @@ Ext4.define('LABKEY.elisa.RunDetailsPanel', {
     },
 
     initComponent : function() {
-
-        this.items = [
-            this.createRunDetailView()
-        ];
+        this.items = [];
 
         this.callParent([arguments]);
+
+        this.queryRunDetails();
     },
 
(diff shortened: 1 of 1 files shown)


labkey-klum:
@cnathe, thanks for doing this! We are planning to do the development for the ELISA run details enhancements in 20.7 due to the time sensitivity for the lab. Do you think it makes sense to target this for 20.7 instead? Otherwise I would imagine some merge issues going from 20.7 -> develop

cnathe:
> @cnathe, thanks for doing this! We are planning to do the development for the ELISA run details enhancements in 20.7 due to the time sensitivity for the lab. Do you think it makes sense to target this for 20.7 instead? Otherwise I would imagine some merge issues going from 20.7 -> develop

Ah yes, good catch. I'll close out this PR and open a new one with a new fb name for release20.7-SNAPSHOT. Thanks.

PR #240, opened 2020-10-02
ELISA assay run details - fix issue to restore display of run details information panel above plot

cnathe:
#### Rationale
Issue [41521](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=41521): ELISA: run details page no longer showing top panel for run param information

As we are about to update the ELISA assay run details page, I thought it would be good to fix the current issue that was preventing the run details information panel from showing on the current run details page. This way we know where we are actually starting from before making changes.

#### Changes
* simplify the data loading logic for the Ext4 store
* fix JS error with Ext.util.Format.number instead of Ext4.util.Format.number

### DIFF
--- a/elisa/webapp/elisa/runDetailsPanel.js
+++ b/elisa/webapp/elisa/runDetailsPanel.js
@@ -22,15 +22,29 @@ Ext4.define('LABKEY.elisa.RunDetailsPanel', {
     },
 
     initComponent : function() {
-
-        this.items = [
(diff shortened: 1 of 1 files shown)


labkey-klum (approved):

cnathe:
Almost forgot that a merge to 20.7 should have a triage approved issue open for it. Let me get that created now and a link added to this PR.

PR #245, opened 2020-10-12
20.7 fb elisa vis 7808

labkey-klum:
#### Rationale
This is a follow on story to the initial [ELISA assay improvements](https://docs.google.com/document/d/131Mh8D1HMs89YltSr263w931cbsqBi96B--xmTVu4OU/edit) which added support for multi-plate, multi-well data formats. In this story the run details page is reworked to support additional filtering capabilities on:
- plate
- analyte
- samples
- controls

The previous ExtJS based view has been replaced with a new React based view.

#### Related Pull Requests
https://github.com/LabKey/platform/pull/1628

#### Changes
- allow passing XY pairs of plot data to the visualization API
- React based run details view for the ELISA assay
- store curve fit parameters as a JSON text string in a new protocol schema table : CurveFit


elisa/resources/schemas/dbscripts/postgresql/elisa-0.000-20.000.sql
```diff
@@ -0,0 +1,41 @@
...
+    RunId INTEGER NOT NULL,
+    ProtocolId INTEGER NOT NULL,
+    PlateName VARCHAR(250),
+    Spot INTEGER,
+    RSquared DOUBLE PRECISION,
+    FitParameters VARCHAR(500),
+
+    CONSTRAINT PK_CurveFit PRIMARY KEY (RowId),
+    CONSTRAINT FK_CurveFit_ProtocolId FOREIGN KEY (ProtocolId) REFERENCES exp.Protocol (RowId),
+    CONSTRAINT FK_CurveFit_RunId FOREIGN KEY (RunId)
```
cnathe, line 36: Any reason to CREATE INDEX for either of these FKs?

elisa/resources/schemas/elisa.xml
```diff
@@ -0,0 +1,42 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+
+ -->
+<tables xsi:schemaLocation="http://labkey.org/data/xml ../../../../../schemas/tableInfo.xsd"
+        xmlns="http://labkey.org/data/xml" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <table tableName="CurveFit" tableDbType="TABLE">
+        <description>Contains one row per notebook or experiment in this folder.</description>
```
cnathe: looks like this might be a copy-paste error from a different xml file.

cnathe (approved):
Approved server side code changes. Pending review of client side code from Marty.

elisa/src/org/labkey/elisa/ElisaProviderSchema.java
```diff
@@ -17,25 +17,6 @@
 {
     public static final String CURVE_FIT_METHOD_TABLE_NAME = "CurveFitMethod";
 
-    enum ElisaCurveFits
-    {
-        FOUR_PARAMETER(StatsService.CurveFitType.FOUR_PARAMETER.getLabel()),
-        FIVE_PARAMETER(StatsService.CurveFitType.FIVE_PARAMETER.getLabel()),
```
cnathe, line 23: @labkey-klum  Any reason we need/want to leave the 5PL type around for the single plate assay case? I believe it was their prior to this work. I have a local 20.7 MSSQL DB which has a run that had Curve Fit Method = Five Parameter. The upgraded/populated CurveFit table for that container looks like this:
Screen Shot 2020-10-15 at 11 47 41 AM

elisa/resources/schemas/dbscripts/postgresql/elisa-0.000-20.000.sql
labkey-klum: yup, good point

elisa/src/org/labkey/elisa/ElisaProviderSchema.java
labkey-klum: I decided to remove it because it didn't yield good fits. I think if we wanted to add a 5PL fit in the future we would use the new 4PL simplex algorithm.

cnathe: That sounds good. Just checking. Thanks.

elisa/test/styleMock.ts
```diff
@@ -0,0 +1,3 @@
+
+
+export {};
```
labkey-martyp, line 3: Is this file necessary?

elisa/src/client/RunDetails/components/PlotButtonBar.tsx
```diff
@@ -0,0 +1,59 @@
...
+    runId: number,
+    runPropertiesRow: {[key: string]: any},
+    plotOptions: PlotOptions,
+    plotId: string
+}
+
+export class PlotButtonBar extends PureComponent<Props> {
+
+    getSVGElement() {
+        return $('#' + this.props.plotId + '>svg');
```
labkey-martyp: Like mentioned above, wonder if this could be done with React.createRef instead of using jQuery

elisa/src/client/RunDetails/components/CurveFitPanel.tsx
```diff
@@ -0,0 +1,64 @@
...
+                            <input
+                                type="checkbox"
+                                name='showCurve'
+                                checked={plotOptions.showCurve}
+                                onChange={this.onShowCurveCheck}
+                            />
+                            Show curve fit line
+                            <LabelHelpTip
+                                title={'Standard Curve Fit'}
+                                body={() => (
```
labkey-martyp: Looks like LabelHelpTip is a PureComponent.  By redeclaring the function here with every render it's causing the body prop to always be different and unnecessary re-renders of LabelHelpTip

elisa/src/client/RunDetails/components/CalibrationCurvePanel.tsx
```diff
@@ -0,0 +1,83 @@
...
+        return $('.plot-panel-section').width();
+    }
+
+    renderPlot() {
+        const { runPropertiesRow, plotOptions, data, curveFitData, columnInfo } = this.props;
+        const LABKEY = getServerContext();
+
+        this.getPlotElement().html('<div class="loading-msg"><i class="fa fa-spinner fa-pulse"></i> Rendering plot...</div>');
+
+        setTimeout(() => {
```
labkey-martyp, line 49: This setTimeout seems odd. Is it a re-render issue?

elisa/src/client/RunDetails/components/CalibrationCurvePanel.tsx
```diff
@@ -0,0 +1,83 @@
...
+        this.renderPlot();
+    }
+
+    getPlotElement() {
+        return $('#' + this.state.plotId);
+    }
+
+    getPlotWidth(): number {
+        return $('.plot-panel-section').width();
+    }
```
labkey-martyp: Wonder if you could use React Refs and not need the dependency on jQuery

elisa/src/client/RunDetails/actions.ts
```diff
@@ -0,0 +1,38 @@
...
+                numberOfPoints: 10000
+            },
+            success: Utils.getCallbackWrapper((data) => resolve(data)),
+            failure: Utils.getCallbackWrapper((error) => reject(error))
+        });
+    });
+}
+
+export function exportSVGToFile(svgEl, format, title: string) {
+    if (svgEl && svgEl.length > 0) {
```
labkey-martyp: svgEl?.length > 0

elisa/src/client/RunDetails/RunDetails.tsx
```diff
@@ -0,0 +1,245 @@
...
+                    <Alert bsStyle={'warning'}>
+                        Warning: the assay design is missing the following fields which may affect the various plotting
+                        features and display of this page: {missingFieldKeys.join(', ')}.
+                    </Alert>
+                }
+                <div className={'row'}>
+                    <div className={'col col-xs-12 col-md-3 run-details-left'}>
+                        <DataSelectionsPanel {...this.state} setPlotOption={this.setPlotOption}/>
+                        <PlotOptionsPanel {...this.props} {...this.state} setPlotOption={this.setPlotOption}/>
+                        <CurveFitPanel {...this.props} {...this.state} setPlotOption={this.setPlotOption}/>
```
labkey-martyp, line 237: Along the lines of the comment above, you might be be able to factor out some common interfaces used as props/state for these components and RunDetails.

elisa/src/client/RunDetails/RunDetails.tsx
```diff
@@ -0,0 +1,245 @@
...
+        return (
+            <>
+                {missingFieldKeys?.length > 0 &&
+                    <Alert bsStyle={'warning'}>
+                        Warning: the assay design is missing the following fields which may affect the various plotting
+                        features and display of this page: {missingFieldKeys.join(', ')}.
+                    </Alert>
+                }
+                <div className={'row'}>
+                    <div className={'col col-xs-12 col-md-3 run-details-left'}>
```
labkey-martyp: I might be misunderstanding the .col class but thought that was more for when you don't have a specific col size class (col-xs-12) on a column.

elisa/src/client/RunDetails/RunDetails.tsx
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    componentDidUpdate(prevProps: Readonly<ImplProps>, prevState: Readonly<ImplState>, snapshot?: any): void
+    {
+        if (shouldReloadCurveFitData(prevState.plotOptions, this.state.plotOptions)) {
+            this.getCurveFitData();
+        }
+    }
+
+    getCurveFitData() {
```
labkey-martyp: arrow function and return type?

elisa/src/client/RunDetails/RunDetails.tsx
```diff
@@ -0,0 +1,245 @@
...
+}
+
+interface ImplProps {
+    protocolId: number,
+    runId: number,
+    runPropertiesRow: {[key: string]: any},
+    data: any[],
+    columnInfo: {[key: string]: any},
+    measures: string[]
+}
```
labkey-martyp: You could carve out a common interface or type between these props and the state for the RunDetails.  Removes some duplicate code and makes sure they're in sync.

elisa/src/client/RunDetails/RunDetails.tsx
```diff
@@ -0,0 +1,245 @@
...
+                this.setState(() => ({ dataError: reason.exception }));
+            }
+        });
+    }
+
+    render() {
+        const { runPropertiesError, dataError, runPropertiesRow, data } = this.state;
+
+        if (runPropertiesError || dataError) {
+            return <Alert>{runPropertiesError || dataError}</Alert>
```
labkey-martyp: Not sure if possible on the server-side but looking at client side code looks like there could be both of these errors.  Only show one?

elisa/src/client/RunDetails/RunDetails.tsx
```diff
@@ -0,0 +1,245 @@
...
+                }
+            },
+            failure: (reason) => {
+                console.error(reason);
+                this.setState(() => ({ runPropertiesError: reason.exception }));
+            }
+        });
+    }
+
+    queryData() {
```
labkey-martyp: This is called in a callback in a callback, seems like it should be arrow function to ensure scope.  And missing return type.

labkey-martyp (commented):
Looks good Cory.  A handful of comments and questions below.  I did not test this out, comments just based on looking over the code. One general comment, there are a lot of missing return types on functions.

elisa/test/styleMock.ts
cnathe: Not at this time, but I decided to keep it in to be consistent with other module jest setups.

elisa/src/client/RunDetails/components/CurveFitPanel.tsx
cnathe: done

elisa/src/client/RunDetails/components/PlotButtonBar.tsx
cnathe: Good call, done. We should likely switch the biologics usage like this at some point as well.

elisa/src/client/RunDetails/components/CalibrationCurvePanel.tsx
cnathe: It is just to help the plot refresh slow down a bit and look more natural so that it briefly show the user the "Rendering plot..." message. Without it, the plot just sort of flashes and is updated, so this looks better in my opinion.

elisa/src/client/RunDetails/components/CalibrationCurvePanel.tsx
cnathe: Yes, done.

elisa/src/client/RunDetails/actions.ts
cnathe: With the switch to React.createRef(), this is no longer an array so the if statement changed to just a null check.

elisa/src/client/RunDetails/RunDetails.tsx
cnathe: You are right, the 'col' isn't needed here. Removed.

elisa/src/client/RunDetails/RunDetails.tsx
cnathe: It doesn't need to be an arrow function as it is only called from componentDidMount and componentDidUpdate. I'll add the void return type.

elisa/src/client/RunDetails/RunDetails.tsx
cnathe: I split out a CommonRunProps interface to extend in a few places.

elisa/src/client/RunDetails/RunDetails.tsx
cnathe: I factored out a CommonRunProps interface to extend in a few places.

elisa/src/client/RunDetails/RunDetails.tsx
cnathe: Done

elisa/src/client/RunDetails/RunDetails.tsx
cnathe: Done

labkey-martyp (approved):

PR #247, opened 2020-10-14
Item 7959: NAb assay updates to for more data access from NAbSpecimen table (and copied-to-study dataset)

cnathe:
#### Rationale
Ticket [41023](https://www.labkey.org/Atlas/support%20tickets/issues-details.view?issueId=41023): Adding full curve NAb data to export
Feature Request [41520](https://www.labkey.org/Dataspace/Feature%20Requests/issues-details.view?issueId=41520): NAb - dilution data improvements

The HVTN lab would like to be able to export all of the fields associated with the dilutions curves from NAb data from the study folders in addition to their normal “ldo_export” view, explained in ticket 41023, so that the dilution data can be submitted to the SCHARP stats team. They would like to incorporate it as a standard dataset available after the NAb data is copied to study.

Additionally, the DataSpace team has been pulling the dilution data, as well as the fit parameter data, for the CAVD DataSpace portal. Adding dilution data, cell/virus control aggregates, and fit parameter data to be accessible from the copied-to-study dataset would also benefit them as they could simplify their current process that uses the selectRows API against the study dataset and then uses the getNAbRuns API against the assay folder in a multi step process.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1630

#### Changes
* nab schema SQL upgrade script to add FitParameters column to NAbSpecimen
* deferred Java upgrade script to queue up root pipeline job to update new FitParameters field for existing NAb protocols/runs
* NabManager helper to ensureFitParamters, updating the NAbSpecimen row if value not already set
* NabPopulateFitParametersPipelineJob to iterate through all NAbAssayProviders to get all protocol runs and set fit parameters for those which are null, logging how many rows were updated / already complete for each NAb protocol
* Add two virtual tables for cell and virus control aggregate values per run with aliased columns via the Runs table 
* Add aliased column for NAb results (i.a. NAbSpecimen) table to FK to DilutionData table

### DIFF
--- a/nab/src/org/labkey/nab/NabManager.java
+++ b/nab/src/org/labkey/nab/NabManager.java
@@ -63,6 +63,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import static org.labkey.api.assay.dilution.DilutionDataHandler.FIT_PARAMETERS_PROPERTY_NAME;
+
 /**
  * User: brittp
  * Date: Oct 26, 2006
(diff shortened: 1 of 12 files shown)


nab/src/org/labkey/nab/query/NabControlAggregatesTable.java
```diff
@@ -0,0 +1,99 @@
...
+    private final String _controlName;
+
+    public NabControlAggregatesTable(String tableName, String controlName, NabProtocolSchema schema, ContainerFilter cf, ExpProtocol protocol, AssayProvider provider)
+    {
+        super(schema.getDbSchema(), tableName, schema, cf);
+        _controlName = controlName;
+        setDescription("Contains one row per run for the aggregate values of the control well data values.");
+
+        var runCol = new BaseColumnInfo("RunId", this, JdbcType.INTEGER);
+        runCol.setFk(new LookupForeignKey(getContainerFilter(), "RowID", null)
```
labkey-klum: I always forget about this, but consider in the future using the variant of setFk which takes a `Builder<ForeignKey>` argument. Matt introduced this with the mutable ColumnInfo changes and its pretty nifty and can yield a more concise syntax for FK creation.

labkey-klum: Ah, I see you already know about this pattern 👍

nab/src/org/labkey/nab/NabManager.java
```diff
@@ -204,4 +215,26 @@ public void getDataPropertiesFromNabRunData(NabRunDataTable nabRunDataTable, Str
...
+     * Ensure the fit parameter value (as a JSON object) for the given NAb assay run specimen results row.
+     * If not already saved to the DB for the given NAbSpecimen row, save it.
+     * @param user
+     * @param assayRun
+     * @param specimenRow
+     * @param dilutionSummary
+     * @return CurveFit.Parameters as Map
+     * @throws FitFailedException
+     */
+    public Map<String, Object> ensureFitParameters(User user, NabSpecimen specimenRow, @Nullable DilutionAssayRun assayRun, @Nullable DilutionSummary dilutionSummary) throws FitFailedException
```
labkey-klum, line 230: Should we consider deprecating this function with a note that is mostly needed in the upgrade case and in 23.3.0 we can delete this since all upgrade cases that we support will have already been run?

labkey-klum (approved):

nab/src/org/labkey/nab/NabManager.java
cnathe: Yes, I'll add that note / tag.

nab/src/org/labkey/nab/query/NabControlAggregatesTable.java
cnathe: Good catch. I was just copying that run column FK from another NAb table, but have switched it to `runCol.setFk(QueryForeignKey.from(schema, cf).to("Runs", "RowID", null));`

PR #258, opened 2020-11-12
Issue 41807: Luminex query performance issues

labkey-jeckels:
#### Rationale
Improve our SQL generation for Luminex queries, resulting in 100x or better execution times on large databases

#### Changes
* Avoid double-filtering on many tables. Previously we applied separate conditions for the container and the protocol (assay design). Collapse to a single filter that handles both.

* No need to filter analytetitration via both titration and analyte. Titration has a more direct link to the run table, so use that exclusively

* A bit of general code cleanup to remove deprecation warnings

### DIFF
--- a/luminex/src/org/labkey/luminex/query/AbstractLuminexTable.java
+++ b/luminex/src/org/labkey/luminex/query/AbstractLuminexTable.java
@@ -15,7 +15,6 @@
  */
 package org.labkey.luminex.query;
 
-import org.labkey.api.data.Container;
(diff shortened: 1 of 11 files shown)


labkey-jeckels:
@cnathe I'm not yet sure which version we'll need to patch, but wanted to get TeamCity running tests and initiate a code review.

cnathe (approved):
Code changes look good. 100x improvement sounds awesome!

PR #293, opened 2021-01-15
Issue 42197: flow: metrics for imported run types

labkey-kevink:
#### Rationale
Add flow folder metrics to count flow data types and how the data was imported

### DIFF
--- a/flow/src/org/labkey/flow/controllers/FlowController.java
+++ b/flow/src/org/labkey/flow/controllers/FlowController.java
@@ -16,25 +16,18 @@
 package org.labkey.flow.controllers;
 
 import org.apache.commons.lang3.StringUtils;
-import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.LogManager;
-import org.labkey.api.action.FormHandlerAction;
+import org.apache.logging.log4j.Logger;
 import org.labkey.api.action.FormViewAction;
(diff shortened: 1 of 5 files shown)


labkey-kevink:
Here's an example of the generated json

[flow-metrics.txt](https://github.com/LabKey/commonAssays/files/5820901/flow-metrics.txt)

flow/src/org/labkey/flow/controllers/FlowController.java
```diff
@@ -412,6 +391,23 @@ public void addNavTrail(NavTree root)
         }
     }
 
+    @RequiresPermission(AdminPermission.class)
+    public class MetricsAction extends ReadOnlyApiAction<Object>
```
labkey-jeckels, line 395: You may know this already, but you can get a preview of the full JSON metrics via Admin Console->Site Settings as well. No problems with having this action available too.

labkey-jeckels (approved):
This looks fine. Do you anticipate use cases for viewing metrics where you don't start at the root container and include everything? Doing it on a per-folder basis and summing will be slower than just a simple unfiltered count.

labkey-kevink:
No, I don't expect to get metrics for a single container.  It was just easier to write the queries using LabKey SQL and the flow schema doesn't work with cross container queries.  If gathering the metrics is slow, I could rewrite the queries in database sql.

PR #296, opened 2021-01-29
Issue 42355: Perf problem during copy-to-study with many rows

labkey-jeckels:
#### Rationale
Users want to be able to copy large assay runs to study datasets. It's painfully slow because we are iterating on each row and doing lots of DB queries

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/1945

#### Changes
* Switch AssayProvider to encourage implementations to query rows in bulk
* Cache ExpData, ExpRun, and ExpExperiment references during process as they're likely to be shared by result rows

### DIFF
--- a/flow/src/org/labkey/flow/data/FlowAssayProvider.java
+++ b/flow/src/org/labkey/flow/data/FlowAssayProvider.java
@@ -21,7 +21,6 @@ import org.jetbrains.annotations.Nullable;
 import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerFilter;
 import org.labkey.api.data.ContainerManager;
-import org.labkey.api.data.ConvertHelper;
 import org.labkey.api.data.DataRegion;
 import org.labkey.api.data.TableSelector;
 import org.labkey.api.exp.OntologyManager;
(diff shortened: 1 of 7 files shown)


flow/src/org/labkey/flow/data/FlowAssayProvider.java
```diff
@@ -500,27 +504,33 @@ public DataExchangeHandler createDataExchangeHandler()
...
+        for (Integer dataRowId : rowIds)
+        {
+            Container container = null;
+            ExpData data = getDataForDataRow(dataRowId, protocol);
+            if (data != null)
+            {
+                ExpRun run = data.getRun();
+                if (run != null)
+                {
+                    Map<String, Object> properties = OntologyManager.getProperties(run.getContainer(), run.getLSID());
```
labkey-kevink, line 519: Maybe cache the targetStudyId using the run rowId as the key to avoid getting the run object and getting the ontology properties for it for each of the result rows.

labkey-kevink (approved):

flow/src/org/labkey/flow/data/FlowAssayProvider.java
labkey-jeckels: Happily, OntologyManager.getProperties() does its own caching so I don't think an extra layer here is necessary.

labkey-klum (approved):

PR #305, opened 2021-03-03
Issue 1713: Suppress trailing ? in URL strings if no parameters

labkey-kevink:
don't use string concatenation to build URLs


labkey-adam (approved):

labkey-adam:
I mean, obviously you need to remove the three old lines before merging. JSP won't compile currently.

PR #306, opened 2021-03-05
42596: IllegalStateException copying Elispot assay data to study

labkey-klum:
#### Rationale
Issue was introduced during a refactor, which resulted in an error always being thrown even if there was data for the run ID.

#### Related Issue
- https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42596


labkey-jeckels (approved):
Thanks for the fix.

Instead of tracking with a boolean, you could also check if result.isEmpty()

labkey-klum:
yes I did consider that but I wasn't sure if the contract of the method was to fail if we returned no data for any of the runs passed in or only if we found no data for any of the runs. The previous implementation handled only a single run at a time.

labkey-jeckels:
Ah, I missed the outer for() loop. Your approach seems good. I don't think we have a formal contract defined but if we can't find a match for every run, there's probably something very wrong.

PR #312, opened 2021-03-16
Fb sample type dataset

labkey-klum:
#### Rationale
Minor updates that were part of the larger refactor in platform to support sample datasets in study.

#### Related Pull Requests
https://github.com/LabKey/platform/pull/2099


RosalineP (commented):
Looks like there are some test outages that might be legitimate on `nab`, or might just be branch drift. I've gone ahead and refreshed all three branches against `develop` and TC is chewing on the changes, let's see if that clears up the red `nab` tests

RosalineP (approved):

PR #326, opened 2021-04-21
Issue 42990: Allow the NabPopulateFitParametersPipelineJob to be rerun to attempt to populate NAbSpecimen FitParameters for previously failed runs

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=42990
The original NabPopulateFitParametersPipelineJob run as part of the NAb Java upgrade script in 20.11 failed on the client server for a couple of the assay protocols. This PR adds a try/catch around the run level data population of that pipeline job so that a single run failure doesn't prevent the data from being populated for the rest of the assay protocol. This PR also registers a new special system maintenance task as a way to re-queue this pipeline job manually.

#### Related Pull Requests
* https://github.com/LabKey/commonAssays/pull/326
* https://github.com/LabKey/platform/pull/2206

#### Changes
* add new NAbPopulateFitParametersTask to re-queue the NabPopulateFitParametersPipelineJob
* update NabPopulateFitParametersPipelineJob to add try/catch for the assay run data access exception handling

### DIFF
--- a/nab/src/org/labkey/nab/NAbPopulateFitParametersTask.java
+++ b/nab/src/org/labkey/nab/NAbPopulateFitParametersTask.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2017 LabKey Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.labkey.nab;
+
+import org.apache.logging.log4j.Logger;
+import org.labkey.api.data.ContainerManager;
+import org.labkey.api.pipeline.PipeRoot;
+import org.labkey.api.pipeline.PipelineService;
+import org.labkey.api.util.SystemMaintenance.MaintenanceTask;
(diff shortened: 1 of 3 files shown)


nab/src/org/labkey/nab/NAbPopulateFitParametersTask.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Override
+    public void run(Logger log)
+    {
+        try
+        {
+            ViewBackgroundInfo info = new ViewBackgroundInfo(ContainerManager.getRoot(), null, null);
+            PipeRoot root = PipelineService.get().findPipelineRoot(ContainerManager.getRoot());
+            PipelineService.get().queueJob(new NabPopulateFitParametersPipelineJob(info, root));
```
labkey-adam, line 53: The MaintenanceTask is already running in a pipeline job... so is there a reason to queue up another job here? Would it make sense to just invoke the NabPopulateFitParametersPipelineJob synchronously?

cnathe: @labkey-adam I had it working that way, as I agree that it is a little odd to have a System Maintenance job where the only thing in its log is two lines that say it started another pipeline job. However, when the NAb fit param job runs within that System Maintenance job, the description of the job just says "System Maintenance" and the file name is something like "system_maintenance_2021-04-22_01-59-59.log" instead of it having information that would point to it being a NAb fit param job. I'm just envisioning the scenario where we ask the client "Can you send us the log file from that NAb fit parameter pipeline job" and them not being able to find it from the data pipeline grid view without having to look through each of the system_maintenance....log files. So it can definitely work either way, thoughts?

labkey-adam: Agreed, either way works and your approach seems fine, given the goals.

I think the NAb parameter fit pipeline job log will end up in the root as well, which might not be super discoverable.

labkey-klum (approved):

PR #336, opened 2021-05-20
Remove Protein Expression Matrix assay

labkey-martyp:
#### Rationale
Per Josh's request, this assay is not being used so we can remove it.  Removed all Protein Expression Matrix code from MS2 module.

#### Changes
* Removed all Protein Expression Matrix code from MS2 module.


labkey-klum:
Unless we are planning to delete the module soon, I would vote to delete the implementation classes as well. It would be less code to build and maintain from refactors.

labkey-jeckels (changes requested):
I agree with Karl. The MS2 module may get removed from the community edition, but it's not going to be totally deleted.

I hope that it's pretty painless to delete everything in the org.labkey.ms2.matrix package. I only see a few references to it, all in MS2Module where it registers implementations, including the assay provide that you've already deleted in this feature branch

labkey-jeckels (approved):

labkey-klum (approved):

PR #342, opened 2021-06-14
fix Ext3 dependency

labkey-matthewb:
#### Rationale
Fix after removing clientapi/ext3 dependency from DataView

### DIFF
--- a/ms2/src/org/labkey/ms2/protein/uploadCustomProteinAnnotations.jsp
+++ b/ms2/src/org/labkey/ms2/protein/uploadCustomProteinAnnotations.jsp
@@ -90,9 +90,9 @@
             <td width="100%">
                 <textarea id="annotationsText" style="width: 100%" rows="15" cols="50" name="annotationsText"><%= h(bean.getAnnotationsText()) %></textarea>
                 <script type="text/javascript">
-                    LABKEY.requiresExt3ClientAPI(function() {
+                    <labkey:loadClientDependencies>
                         Ext.EventManager.on('annotationsText', 'keydown', LABKEY.ext.Utils.handleTabsInTextArea);
-                    });
+                    </labkey:loadClientDependencies>
                 </script>
             </td>
         </tr>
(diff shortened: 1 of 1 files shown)


ms2/src/org/labkey/ms2/protein/uploadCustomProteinAnnotations.jsp
```diff
@@ -90,7 +90,9 @@
             <td width="100%">
                 <textarea id="annotationsText" style="width: 100%" rows="15" cols="50" name="annotationsText"><%= h(bean.getAnnotationsText()) %></textarea>
                 <script type="text/javascript">
-                    Ext.EventManager.on('annotationsText', 'keydown', LABKEY.ext.Utils.handleTabsInTextArea);
+                    LABKEY.requiresExt3ClientAPI(function() {
```
labkey-nicka: I assume this is being caused by some asynchronous loading of this page? I would recommend using `<labkey:loadClientDependencies>` instead. This will respect the client dependencies already declared in the JSP so it doesn't have to be redeclared here. Something like:

```jsp
<script type="text/javascript">
    <labkey:loadClientDependencies>
        Ext.EventManager.on('annotationsText', 'keydown', LABKEY.ext.Utils.handleTabsInTextArea);
    </labkey:loadClientDependencies>
</script>
```

labkey-matthewb: This one might be fine, I was just fixing all the files that had this same pattern.  I think the problem with some of the other usages in the other platform PR were because the jsp's were constructed/rendered late, so their dependencies were not being added to the page.

labkey-nicka: Yup, this is also a case that `<labkey:loadClientDependencies/>` was intended to fix.

labkey-matthewb: I see, avoids having to declare the dependencies twice.  Clever.

labkey-nicka (approved):

PR #353, opened 2021-07-26
Drop support for SQLServer in Luminex module

labkey-jeckels:
#### Rationale
We don't have users for this module on SQL Server, and it's been causing test failures on TeamCity on SQL Server agents

#### Changes
* Remove SQL Server script
* Declare Postgres only in module and test code


labkey-jeckels:
@labkey-sweta we can drop this TeamCity project as part of this change:

https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_ModuleSuites_LuminexSqlserve?branch=&mode=builds

labkey-sweta (approved):

PR #385, opened 2021-11-16
Use helper methods to wait for and switch browser windows

labkey-tchad:
#### Rationale
Tests fail intermittently when the new window doesn't open quite fast enough.

#### Changes
* Use helper methods to wait for and switch browser windows


labkey-chrisj (approved):
ahhh.  better

PR #388, opened 2021-11-30
Issue 44391: MS2 tables fail to propagate ContainerFilter across lookups

labkey-jeckels:
#### Rationale
The big ContainerFilter refactor left some TODOs behind in MS2 where they're not propagated through lookups. We need to make them work.

#### Changes
* Push the right ContainerFilter through lookups and into table constructors


labkey-matthewb (changes requested):
If we're touching these lookups (yeah!), it would be great to convert to QueryForeignKey if there is not good reason not to.  Also QueryForeignKey.Builder is a readable option.

labkey-jeckels:
@labkey-matthewb sadly all but one case isn't well-suited to QFK. Either the lookup target isn't exposed as a UserSchema table, or we're configuring the table and/or URL for the lookup column.

I pushed the one change that didn't require a bigger refactor of either QFK or the MS2 schema and tables.

labkey-matthewb (approved):

PR #396, opened 2021-12-15
44444 FlowProtocolSchema.createDataTable()

labkey-matthewb:
#### Rationale
FlowProtocolSchema can use FlowSchema.getTable() to implement createDataTable().  This results in a more consistently constructed TableInfo.

#### Related Pull Requests
* 

#### Changes
*


RosalineP (approved):
Thanks for the update!

PR #414, opened 2022-01-07
fix lookups that use internal create{Table}() methods.

labkey-matthewb:
#### Rationale
Flow lookups that use internal createTable() methods need to be updated to be properly constructed (They need to call afterConstruct()).  Unfortunately, these can't be directly migrated to use FlowSchema.getTable().

#### Related Pull Requests
https://github.com/LabKey/commonAssays/pull/396

#### Changes
*


labkey-jeckels (approved):

labkey-jeckels:
Thanks for the comments to help us make sense of this in the future.

PR #419, opened 2022-01-16
Remove obsolete upgrade code

labkey-adam:
#### Rationale
This upgrade code no-ops at bootstrap time, the only case where it's called


nab/src/org/labkey/nab/NabUpgradeCode.java
```diff
@@ -1,46 +0,0 @@
...
-        {
-            try
-            {
-                ViewBackgroundInfo info = new ViewBackgroundInfo(ContainerManager.getRoot(), null, null);
-                PipeRoot root = PipelineService.get().findPipelineRoot(ContainerManager.getRoot());
-
-                // Create a new pipeline job and add it to the JobStore. Since this code runs before the module
-                // startBackgroundThreads(), the PipelineModule.startBackgroundThreads JobRestarter will pick up this
-                // new job and queue/run it.
-                NabPopulateFitParametersPipelineJob job = new NabPopulateFitParametersPipelineJob(info, root);
```
labkey-klum, line 36: I wonder if we can delete the pipeline job and the maintenance task associated with this upgrade. @cnathe do you remember if the maintenance task was just a way to expose the upgrade and whether it's needed anymore?

labkey-klum (approved):

nab/src/org/labkey/nab/NabUpgradeCode.java
labkey-adam: Right, I see that it's the only maintenance task that overrides `isRecurring()` to return false. So that would seem to indicate a one-time invocation option.

For the future, note that we now have a way to invoke any Java upgrade code method from the server UI (it hangs off the admin console SQL Scripts page). So, the one-off actions and maintenance tasks should no longer be needed.

cnathe: We had add this `NAbPopulateFitParametersTask` as a system maintenance task because we wanted a way for the Atlas team to manually re-trigger the job as needed. That is, we expected there are to some NAb assay runs that failed to populate the fit params the first time it ran (i.e. via the upgrade code invocation) but then if they fixed things up and wanted to run it again manually, we provided this mechanism.

I believe that Atlas is done with this task and has completed their upgrade, so I think you are right @labkey-klum that we could likely remove that `NAbPopulateFitParametersTask` and `NabPopulateFitParametersPipelineJob` at this time.

labkey-adam: Thanks, @cnathe ! I'll remove them.

cnathe (approved):

PR #434, opened 2022-02-22
Luminex ManageGuideSetPanel update to add null check for grid store before trying to load data into it

cnathe:
#### Rationale
Some intermittent TeamCity test failiures with the LuminexGuideSetTesting test cases show a client side error: TypeError: this.allRunsGrid.getStore() is null. This PR adds a check for the grid store to the if statement that handles loading the data into that store.

#### Changes
* Luminex ManageGuideSetPanel update to add null check for grid store before trying to load data into it


cnathe:
The LuminexGuideSetDisablingTest has passed 6 times on this FB on TC:
https://teamcity.labkey.org/project.html?projectId=LabKey_Trunk&branch_LabKey_Trunk=luminexGuideSetTestFailures

labkey-jeckels (approved):

PR #439, opened 2022-03-12
45023: Unauthorized exception for readonly user - Dataset linked from flow data

labkey-klum:
#### Rationale
Datasets sourced from Flow data weren't respecting `ContextualRoles` when resolving the underlying tables:

https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=45023

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/3133

#### Changes
- `FlowSchema` implements `HasContextualRoles`

### DIFF
--- a/flow/src/org/labkey/flow/data/FlowProtocolSchema.java
+++ b/flow/src/org/labkey/flow/data/FlowProtocolSchema.java
@@ -54,6 +54,7 @@ public class FlowProtocolSchema extends AssayProtocolSchema
     public ExpRunTable createRunsTable(ContainerFilter cf)
     {
         FlowSchema flowSchema = new FlowSchema(getUser(), getContainer());
+        getContextualRoles().forEach(flowSchema::addContextualRole);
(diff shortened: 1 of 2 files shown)


flow/src/org/labkey/flow/data/FlowProtocolSchema.java
```diff
@@ -54,6 +54,7 @@ public FlowAssayProvider getProvider()
     public ExpRunTable createRunsTable(ContainerFilter cf)
     {
         FlowSchema flowSchema = new FlowSchema(getUser(), getContainer());
```
labkey-matthewb: This class looks like it's just a wrapper for FlowSchema.  Maybe it would be clearer to follow a wrapper pattern and construct the FlowSchema once in the FlowProtocolSchema constructer and then delegate addContextualRole() and getContextualRoles().  The idea seems right.

labkey-matthewb (approved):

PR #440, opened 2022-03-15
Close DataLoaders when done with them

labkey-tchad:
#### Rationale
Failing to close DataLoaders is causing some imports to fail on Windows when attempting to clean up temp files.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3131

#### Changes
* Close DataLoaders when done with them
* Use try-with-resources when able


Sigmonia (approved):

labkey-matthewb (approved):
Yes!

PR #445, opened 2022-03-22
Don't use HashSet for column lists

labkey-adam:
#### Rationale
We're getting stricter about requiring stable-ordered column lists. See related PR for details. Fix up some cache names.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3158

### DIFF
--- a/viability/src/org/labkey/viability/ViabilityManager.java
+++ b/viability/src/org/labkey/viability/ViabilityManager.java
@@ -16,14 +16,17 @@
 
 package org.labkey.viability;
 
-import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 import org.junit.After;
 import org.junit.Assert;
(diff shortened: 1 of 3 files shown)


viability/src/org/labkey/viability/ViabilityManager.java
```diff
@@ -525,7 +525,7 @@ public static void deleteAll(List<ExpData> datas, Container c)
                 dataIDs.add(data.getRowId());
 
             TableSelector ts = new TableSelector(ViabilitySchema.getTableInfoResults(),
-                    new HashSet<String>(Arrays.asList("RowID", "ObjectID")),
+                    new LinkedHashSet<>(Arrays.asList("RowID", "ObjectID")),
```
labkey-jeckels: Any reason you're using LinkedHashSet here and PageFlowUtil.set() for the other set of identical columns in this class?

labkey-jeckels (approved):

viability/src/org/labkey/viability/ViabilityManager.java
labkey-adam: Nope, conservation of keystrokes maybe?

PR #448, opened 2022-03-30
Issue 45138: Adjust FolderXarImporterFactory to support unzipped XARs

labkey-jeckels:
#### Rationale
Checking in zipped files (like .xar files) makes it hard to edit and view diffs. We can easily handle importing .xar.xml files in folder archives, even if we always export in the compressed form

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3187

#### Changes
* Check for both .xar and .xar.xml files in folder archives
* Remove deprecated File method variants in favor of Path

### DIFF
--- a/flow/src/org/labkey/flow/script/ScriptXarSource.java
+++ b/flow/src/org/labkey/flow/script/ScriptXarSource.java
@@ -71,15 +71,9 @@ public class ScriptXarSource extends XarSource
     }
 
     @Override
-    public File getRoot()
-    {
-        return _root;
-    }
-
-    @Override
     public Path getRootPath()
     {
-        return null != getRoot() ? getRoot().toPath() : null;
+        return null != _root ? _root.toPath() : null;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


flow/src/org/labkey/flow/script/ScriptXarSource.java
```diff
@@ -94,9 +88,10 @@ public ExperimentArchiveDocument getDocument()
         return _doc;
     }
 
+
     @Override
-    public File getLogFile()
+    public Path getLogFilePath()
     {
-        return _logFile;
+        return _logFile.toPath();
```
labkey-susanh, line 95: Could `_logFile` be null?

labkey-susanh (approved):

flow/src/org/labkey/flow/script/ScriptXarSource.java
labkey-jeckels: It's initialized in the constructor so we should always have a value.

PR #458, opened 2022-04-18
Item 10203: Support sample type and data class renaming

XingY:
#### Rationale
Sample type LSID used to be deterministic for a given container and a specific name. With the change to support sample type rename, sample type lsid is switched to be DBSequence based, which means we don't know a sample type's lsid from its name beforehand. Flow joins to "Samples" sample type, and has been using the pre calculated name based lsid for linking. This PR changes the implementation so the linking can only be done on existing "Samples" type, by querying its lsid from DB, rather than guess it. 

#### Related Pull Requests
* https://github.com/LabKey/labkey-ui-components/pull/808
* https://github.com/LabKey/platform/pull/3256
* https://github.com/LabKey/commonAssays/pull/458
* https://github.com/LabKey/sampleManagement/pull/922
* https://github.com/LabKey/biologics/pull/1255

#### Changes
* change to query lsid from existing sample type with name "Samples", instead of detemining lsid with fixed substitution
* update flow test to create "Samples" sample type before define join fields

### DIFF
--- a/flow/src/org/labkey/flow/data/FlowProtocol.java
+++ b/flow/src/org/labkey/flow/data/FlowProtocol.java
@@ -307,7 +307,7 @@ public class FlowProtocol extends FlowObject<ExpProtocol>
         if (propValue != null)
             return propValue;
 
-        // get sample type lsid for sample with name "Samples"
+        // get lsid for sample type with name "Samples"
         ExpSampleType sampleType = getSampleType(user);
         if (sampleType != null)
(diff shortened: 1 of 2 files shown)


labkey-jeckels (approved):
This change seems like an improvement over the current code. I can't be sure but the one non-test use of getSampleTypeLSID() looks like it'll behave reasonably if it gets a null value back.

labkey-susanh (approved):

PR #461, opened 2022-04-21
script nonce attribute

labkey-matthewb:
#### Rationale
Bulk add script nonce attribute in jsp files. 

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/flow/src/org/labkey/flow/webparts/FlowSummary.jsp
+++ b/flow/src/org/labkey/flow/webparts/FlowSummary.jsp
@@ -115,29 +115,31 @@
     }
 </style>
 
-<script type="text/javascript">
+<script type="text/javascript" nonce="<%=getScriptNonce()%>">
     function createRenderer(detailsURL, countProperty, countLabel)
     {
+        const h = LABKEY.Utils.encodeHtml;
+
         return function (el, json) {
             var html = "<table border='0'>";
             for (var i = 0; i < <%=DISPLAY_MAX_ROWS%> && i < json.rows.length; i++) {
                 var row = json.rows[i];
                 var name = row.Name.value;
                 var url = row.Name.url;
-                var comment = row["Flag/Comment"].value ? " title='" + row["Flag/Comment"].value + "'" : "";
+                var comment = row["Flag/Comment"].value ? " title='" + h(row["Flag/Comment"].value) + "'" : "";
                 var iconCls = row["Flag/Comment"].value ? <%=q(FlagColumnRenderer.flagEnabledCls())%> : "";
 
                 html += "<tr>" +
-                        "<td>" + (iconCls ? "<a" + comment + " href='" + url + "'><i class='" + iconCls + "'/></a>" : "") + "</td>" +
-                        "<td nowrap><a" + comment + " href='" + url + "'>" + name + "</a><td>";
+                        "<td>" + (iconCls ? "<a" + comment + " href='" + h(url) + "'><i class='" + h(iconCls) + "'/></a>" : "") + "</td>" +
+                        "<td nowrap><a" + comment + " href='" + h(url) + "'>" + h(name) + "</a><td>";
                 if (countProperty) {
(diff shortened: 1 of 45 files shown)


flow/src/org/labkey/flow/webparts/FlowSummary.jsp
```diff
@@ -115,29 +115,31 @@
...
+        const h = LABKEY.Utils.encodeHtml;
+
         return function (el, json) {
             var html = "<table border='0'>";
             for (var i = 0; i < <%=DISPLAY_MAX_ROWS%> && i < json.rows.length; i++) {
                 var row = json.rows[i];
                 var name = row.Name.value;
                 var url = row.Name.url;
-                var comment = row["Flag/Comment"].value ? " title='" + row["Flag/Comment"].value + "'" : "";
+                var comment = row["Flag/Comment"].value ? " title='" + h(row["Flag/Comment"].value) + "'" : "";
```
labkey-jeckels, line 129: Shouldn't this get q() too/instead?

flow/src/org/labkey/flow/webparts/FlowSummary.jsp
```diff
@@ -115,29 +115,31 @@
...
                 var iconCls = row["Flag/Comment"].value ? <%=q(FlagColumnRenderer.flagEnabledCls())%> : "";
 
                 html += "<tr>" +
-                        "<td>" + (iconCls ? "<a" + comment + " href='" + url + "'><i class='" + iconCls + "'/></a>" : "") + "</td>" +
-                        "<td nowrap><a" + comment + " href='" + url + "'>" + name + "</a><td>";
+                        "<td>" + (iconCls ? "<a" + comment + " href='" + h(url) + "'><i class='" + h(iconCls) + "'/></a>" : "") + "</td>" +
```
labkey-jeckels, line 133: Shouldn't this get q() too/instead?

flow/src/org/labkey/flow/webparts/FlowSummary.jsp
```diff
@@ -115,29 +115,31 @@
...
                 if (countProperty) {
-                    html += "<td align='right' nowrap>(" + row[countProperty].value + " " + countLabel + ")</td>";
+                    html += "<td align='right' nowrap>(" + h(row[countProperty].value) + " " + h(countLabel) + ")</td>";
                 }
                 html += "</tr>";
             }
             if (json.rows.length > <%=DISPLAY_MAX_ROWS%>) {
                 html += "<tr>";
-                html += "<td colspan=2><a href='" + detailsURL + "'>More...</a></td>";
+                html += "<td colspan=2><a href='" + h(detailsURL) + "'>More...</a></td>";
```
labkey-jeckels, line 142: Shouldn't this get q() too/instead?

labkey-jeckels (approved):

flow/src/org/labkey/flow/webparts/FlowSummary.jsp
labkey-matthewb: No.  LABKEY.Utils.encodeHtml() will remove any single-quote, double-quote, or backslash characters, so I believe this is correct.  In any event, q() is a server-side method.

flow/src/org/labkey/flow/webparts/FlowSummary.jsp
labkey-matthewb: It would be good to write a unit test that ensures the correct behavior or htmlEncode() though.

labkey-matthewb: update: I checked that encodeURI() handles backslash, might need to add that to htmlEncode().  (not relevant to this use case, but good to do)

labkey-jeckels: Ah, OK. I lost track of which h() this was calling.

PR #464, opened 2022-05-10
Item 9548 - Show peak area or quality info in protein sequence view

ankurjuneja:
#### Rationale
This MR adds the heat map legend for intensity and confidence score in protein view.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/3337

#### Changes
* add new script to add the heat map legends on protein view
* new lombok library
* update proteinCoverageMap to reflect the intensity/confidence colors


ms2/src/org/labkey/ms2/Protein.java
```diff
@@ -448,7 +458,7 @@ void addPeptide(Range range, Integer newLevel, int curIdx, int curRowStart, int
                     if (!_forCoverageMapExport)
                     {
                         details += String.format("\nUnmodified: %d", counts.getCountUnmodifiedPeptides());
-                        label = helpPopup("Peptide Details", details, false, linkText, 200, onClickScript );
+                        label = helpPopup("Peptide Details", details, false, "<div style=color:" + range.pepcounts.foregroundColor +">" + linkText + "</div>", 200, onClickScript );
```
labkey-jeckels:
```suggestion
                        label = helpPopup("Peptide Details", details, false, "<div style=\"color:" + range.pepcounts.foregroundColor +"\">" + linkText + "</div>", 200, onClickScript );

```

ms2/src/org/labkey/ms2/Protein.java
```diff
@@ -476,9 +486,18 @@ void addPeptide(Range range, Integer newLevel, int curIdx, int curRowStart, int
...
+            {
+                color = range.pepcounts.intensityColor;
+            }
+            else if (range.pepcounts.isConfidenceView)
+            {
+                color = range.pepcounts.confidenceColor;
+            }
             if ((range.start == curIdx) || (curRowStart == curIdx))
-                td = String.format(baseOutput, cssClass, colsCurrentRow, label);
+                // pedtide color property
```
labkey-jeckels:
```suggestion
                // peptide color property

```

ms2/src/org/labkey/ms2/Protein.java
```diff
@@ -16,31 +16,41 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
 
 import static org.labkey.api.util.PageFlowUtil.helpPopup;
 
+@Log4j2
```
labkey-jeckels: This approach means that we won't get a helpful description in the Admin Console's Loggers page, like we would if using our simple LogHelper.getLogger() utility method

ms2/src/org/labkey/ms2/Protein.java
```diff
@@ -510,6 +541,15 @@ public void setPeptides(String... peptides)
         int countUnmodifiedPeptides;
         Map<String , Integer> countModifications;
         int countInstances;
+        @Getter @Setter String intensityColor;
+        @Getter @Setter Double intensity;
+        @Getter @Setter String confidenceColor;
+        @Getter @Setter Double confidence;
+        @Getter @Setter String foregroundColor;
+        @Getter @Setter boolean isIntensityView;
```
labkey-jeckels: Why is this a peptide-level setting? We'd never be showing intensity for some and confidence for others, would we?

ms2/src/org/labkey/ms2/Protein.java
```diff
@@ -51,7 +61,7 @@
 
     private boolean _forCoverageMapExport = false;
 
-    private static final String PEPTIDE_START_TD="<td class=\"%s\" colspan=%d > %s </td>";
+    private static final String PEPTIDE_START_TD="<td class=\"%s\" %s=%s colspan=%d > %s </td>";
```
labkey-jeckels: Do we always use "style" for the first new substitution here? If so, can we treat it more like the class substitution here? And we should include quotes on the value regardless.

ms2/src/org/labkey/ms2/proteinCoverageMap.jsp
```diff
@@ -15,34 +15,163 @@
...
 <%!
     @Override
     public void addClientDependencies(ClientDependencies dependencies)
     {
         dependencies.add("MS2/ProteinCoverageMap.css");
         dependencies.add("MS2/ProteinCoverageMap.js");
+        dependencies.add("MS2/PeptideIntensityHeatMap.js");
         dependencies.add("util.js");
         dependencies.add("internal/jQuery");
+        dependencies.add("vis/lib/d3-3.5.17.min.js");
```
labkey-jeckels: Switching to depend on visDependencies.lib.xml gets us dev/prod variants which would be nice. It does pull in some other dependencies we don't need but they should be cached so it won't be a big deal.

ms2/src/org/labkey/ms2/proteinCoverageMap.jsp
```diff
@@ -15,34 +15,163 @@
...
+        // assign blue colors from median to last -> lighter to darker
+        for (int i = medianIndex + 1; i < count; i++)
+        {
+
+            var peptideColor = blueGradient.get(i-medianIndex);
+            var peptideCharacteristic = peptideCharacteristics.get(i);
+            if (isIntensityView)
+            {
+                heatMapColorRGB.put(peptideCharacteristics.get(i).getIntensity(), peptideColor);
+                peptideCharacteristic.setIntensityColor("#" + Integer.toHexString(peptideColor.getRGB()).substring(2));
```
labkey-jeckels: Since we only set one color (either based on intensity or confidence), can do we need to have both intensityColor and confidenceColor on the objects? Could we just have a single color?

ms2/webapp/MS2/PeptideIntensityHeatMap.js
```diff
@@ -0,0 +1,43 @@
...
+                    .enter().append('text')
+                    .text((d) => d.toString())
+                    .attr("y", (d,i) => y_axis + (rectWidth*i) + 25)
+                    .attr("x", x_axis + 35);
+
+        },
+
+        changeView: function (runId) {
+            var viewBy = $("#peptide-setting-select").val();
+            window.location = LABKEY.ActionURL.buildURL('targetedms', 'showProtein' , null, {id: runId, viewBy: viewBy});
```
labkey-jeckels: The MS2 module shouldn't have hard-coded URLs to the targetedms controller like this. I think this could likely take the current URL and replace or add the viewBy parameter

labkey-jeckels (changes requested):

ms2/src/org/labkey/ms2/proteinCoverageMap.jsp
ankurjuneja: done

ms2/webapp/MS2/PeptideIntensityHeatMap.js
ankurjuneja: OK, wasn't aware of that. done

ms2/src/org/labkey/ms2/Protein.java
ankurjuneja: removed this approach for now

ms2/src/org/labkey/ms2/Protein.java
ankurjuneja: removed

ms2/src/org/labkey/ms2/Protein.java
ankurjuneja: updated

ms2/src/org/labkey/ms2/proteinCoverageMap.jsp
ankurjuneja: yes, updated

ms2/src/org/labkey/ms2/proteinCoverageMap.jsp
```diff
@@ -15,34 +15,160 @@
...
-<%@ page import="org.apache.commons.lang3.StringUtils" %>
-<%@ page import="org.labkey.api.protein.ProteinFeature" %>
+<%@ page import="java.util.stream.Collectors" %>
 <%@ page extends="org.labkey.api.jsp.JspBase" %>
 <%
     MS2Controller.ProteinViewBean bean = ((JspView<MS2Controller.ProteinViewBean>)HttpView.currentView()).getModelBean();
+    var currentURL = getActionURL();
+    var viewByParam = currentURL.getParameter("viewBy");
+    var isIntensityView = viewByParam == null || viewByParam.equalsIgnoreCase("intensity");
+    var isConfidenceView = viewByParam != null && viewByParam.equalsIgnoreCase("confidenceScore");
```
labkey-jeckels, line 40: We either are showing intensity of confidence, right? What happens if you have a viewBy parameter that doesn't match either?

ms2/webapp/MS2/PeptideIntensityHeatMap.js
```diff
@@ -0,0 +1,45 @@
+if (!LABKEY.ms2) {
+    LABKEY.ms2 = {};
+}
+
+if (!LABKEY.ms2.PeptideIntensityHeatMap) {
```
labkey-jeckels: Optional - This feels more like a legend than a heat map. The sequence view part is closer to being a heat map.

labkey-jeckels (approved):

labkey-jeckels:
@ankurjuneja there are some test updates to make too

ms2/src/org/labkey/ms2/proteinCoverageMap.jsp
ankurjuneja: the old sequence coverage map (the current functionality prior to this) is displayed which also happens in the case when viewBy parameter is not present.

PR #476, opened 2022-06-20
Adjust logarithmic plots

labkey-matthewb:
#### Rationale
Don't constrain data values to 'expected' range on load or compensate.
Extend plot low range for logarithmic plots
see issue 45300

#### Related Pull Requests
* 

#### Changes
*


labkey-hannah:
Was able to check it out and looks pretty good - service team is going to check with other heavy flow users and then I will approve

labkey-hannah (approved):
Looks good to me (and client-approved!)

PR #582, opened 2023-02-15
[MacCoss] PanoramaWeb public search improvements - protein search optimization

labkey-bpatel:
#### Rationale
Add indexes to improve performance for a query that utilizes tables in ms2 module for MacCoss lab.

See Related PR for more details.

#### Related Pull Requests
* https://github.com/LabKey/MacCossLabModules/pull/313

#### Changes
* Add upgrade scripts to add special indexes with 'varchar_pattern_ops' operator


labkey-bpatel:
I will message the merge master so that I can make updates to the schema version and upgrade script for when it gets merged forward to 23.1. 
I could have created scripts with 23.000-23.001 since there are no new scripts in 23.1 and 23.2 but its just odd to see release 23.x scripts in release 22.x, so went this route.

labkey-adam:
> I will message the merge master so that I can make updates to the schema version and upgrade script for when it gets merged forward to 23.1. I could have created scripts with 23.000-23.001 since there are no new scripts in 23.1 and 23.2 but its just odd to see release 23.x scripts in release 22.x, so went this route.

Actually, our standard best practice for this would be to name the script and version for 23.001, even though this is going into the 22.11.x branch. This removes the burden (on the merge master) and potential confusion around renaming scripts. It also prevents scripts from running twice (though not an issue in this case). You'll notice that the 22.11.x branches on a lot of repos are already set to 23.000 or beyond: EHRs, CDS, TargetedMS, etc.

I guess we're sure we don't need these indexes for SQL Server... if so, we should delete that empty script.

labkey-jeckels (approved):

PR #587, opened 2023-03-02
Use PatriciaTrie from Commons Collections 4

labkey-adam:
#### Rationale
We've been using a one-off, 12-year-old implementation of Patricia-Trie (a space-efficient data structure) to validate that FASTA files contain no duplicate protein names. Commons Collections has its own implementation of Patricia-Trie, so switch over to that and drop the old library.

Also a bit of clean up with `@NotNull` annotations, bad `StringUtils` imports, and bad date formats.

### DIFF
--- a/ms2/src/org/labkey/ms2/protein/fasta/FastaValidator.java
+++ b/ms2/src/org/labkey/ms2/protein/fasta/FastaValidator.java
@@ -32,7 +32,7 @@ public class FastaValidator
     {
     }
 
-    //** Determine if any
+    /** Determine if FASTA file has any duplicate protein names **/
     public List<String> validate(File fastaFile)
     {
         List<String> errors = new ArrayList<>();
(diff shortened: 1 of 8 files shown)


labkey-jeckels (approved):

ms2/src/org/labkey/ms2/protein/fasta/FastaValidator.java
```diff
@@ -15,46 +15,36 @@
...
 {
     /** Use a trie to get space-efficient storage */
-    private final PatriciaTrie _proteinNames = new PatriciaTrie<>(ByteArrayKeyAnalyzer.VARIABLE);
+    private final PatriciaTrie<String> _proteinNames = new PatriciaTrie<>();
 
     public FastaValidator()
     {
     }
 
+    //** Determine if any
```
labkey-susanh: If any what?

PR #609, opened 2023-04-18
Issue 47734: SQLException from bad JDBC parameter order viewing MS2 runs

labkey-jeckels:
#### Rationale
We're not reliably ordering the JDBC parameter values when assembling the SQLFragment for FASTA files

#### Changes
* Add the boolean value before we append the container SQLFragment


labkey-adam (approved):
Did this ever work?

labkey-jeckels:
@labkey-adam for some ContainerFilters, probably not. Not all of them use a JDBC parameter so it works fine in many scenarios.

PR #612, opened 2023-05-02
Don't provision "fast flow temp tables" in folders with no flow data

labkey-matthewb:
#### Rationale
This is avoid unintentionally creating thousands of empty temp tables.
Separately we should consider hiding the "flow" schema when the flow module is not enabled.

#### Related Pull Requests
* 

#### Changes
*


labkey-matthewb:
The non-cleanup portion of this is FlowSchema.java from line 1988 onward.

labkey-adam (approved):
Nice clean ups!

PR #617, opened 2023-05-25
Luminex test failures on MSSQL related to wait time for apply guide set modal grid

cnathe:
#### Rationale
A couple Luminex tests have been failing on MSSQL because of the loading of the apply guide set modal grid. When viewed locally, this grid loads very quickly. This PR increases the wait time for that grid to load.

#### Changes
* increase the wait time for apply guide set dialog grid to load

### DIFF
--- a/luminex/test/src/org/labkey/test/util/luminex/LuminexGuideSetHelper.java
+++ b/luminex/test/src/org/labkey/test/util/luminex/LuminexGuideSetHelper.java
@@ -223,7 +223,7 @@ public class LuminexGuideSetHelper
         WebElement selectedRunsGrid = Locator.byClass("selectedRunsGrid").waitForElement(applyGuideSetWindow, WAIT_FOR_JAVASCRIPT);
         for (String network : networks)
         {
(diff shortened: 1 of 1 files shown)


labkey-danield (approved):

labkey-jeckels:
If this is a recurring problem, also consider forcing the DB system maintenance. We've seen this pattern before where SQLServer doesn't update its statistics immediately after importing data and ends up choosing a bad query plan.

https://github.com/LabKey/testAutomation/blob/5290f2e403350e985ba654d1dc94b227efd8cc4c/src/org/labkey/test/tests/nab/NabAssayTest.java#L382

cnathe:
> If this is a recurring problem, also consider forcing the DB system maintenance. We've seen this pattern before where SQLServer doesn't update its statistics immediately after importing data and ends up choosing a bad query plan.
> 
> https://github.com/LabKey/testAutomation/blob/5290f2e403350e985ba654d1dc94b227efd8cc4c/src/org/labkey/test/tests/nab/NabAssayTest.java#L382

Ah, that is a good call, Josh. Thanks. I will try a small feature branch change to give that a try and revert this extra timeout so that we can let TeamCity run with it for a few cycles. Thanks.

PR #620, opened 2023-06-14
Moving entities between projects - Assay Data

XingY:
#### Rationale
Support moving of assay runs between projects/containers

#### Related Pull Requests
- https://github.com/LabKey/labkey-ui-components/pull/1211
- https://github.com/LabKey/labkey-ui-premium/pull/111
- https://github.com/LabKey/platform/pull/4514
- https://github.com/LabKey/commonAssays/pull/620
- https://github.com/LabKey/sampleManagement/pull/1893
- https://github.com/LabKey/biologics/pull/2181

#### Changes
* handle moving of specialty assay table container fields

### DIFF
--- a/elisa/src/org/labkey/elisa/ElisaAssayProvider.java
+++ b/elisa/src/org/labkey/elisa/ElisaAssayProvider.java
@@ -51,6 +51,7 @@ import org.labkey.api.module.ModuleLoader;
 import org.labkey.api.pipeline.PipelineProvider;
 import org.labkey.api.qc.DataExchangeHandler;
 import org.labkey.api.query.FieldKey;
+import org.labkey.api.query.FilteredTable;
 import org.labkey.api.security.User;
 import org.labkey.api.settings.ExperimentalFeatureService;
 import org.labkey.api.study.assay.ParticipantVisitResolverType;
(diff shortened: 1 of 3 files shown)


elisa/src/org/labkey/elisa/ElisaAssayProvider.java
```diff
@@ -430,4 +434,24 @@ public Collection<StatsService.CurveFitType> getCurveFits()
     {
         return List.of(StatsService.CurveFitType.FOUR_PARAMETER_SIMPLEX, StatsService.CurveFitType.LINEAR);
     }
+
+    @Override
+    protected void moveAssayResults(List<ExpRun> runs, ExpProtocol protocol, Container sourceContainer, Container targetContainer, User user, AssayMoveData assayMoveData)
+    {
+        super.moveAssayResults(runs, protocol, sourceContainer, targetContainer, user, assayMoveData); // assay results
```
labkey-susanh, line 442: Should maybe surround with `ensureTransaction` (here and elsewhere).

labkey-susanh (approved):

elisa/src/org/labkey/elisa/ElisaAssayProvider.java
XingY: This is called inside AssayProvider.moveRuns, which is called by ExperimentService.moveAssayRuns, which is inside of ensureTransaction

labkey-susanh: Yes, but it could be called from somewhere else and does do more than one update, so other callers will need to know to transact.

PR #656, opened 2023-10-18
Don't attempt to include non-finite numbers in JSON

labkey-tchad:
#### Rationale
Non-finite numbers aren't allowed in `JSONObject`s

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/4837

#### Changes
* Prevent non-finite numbers in JSON values

### DIFF
--- a/nab/src/org/labkey/nab/NabRunPropertyMap.java
+++ b/nab/src/org/labkey/nab/NabRunPropertyMap.java
@@ -19,14 +19,15 @@ import org.json.JSONObject;
 import org.labkey.api.assay.dilution.DilutionAssayRun;
 import org.labkey.api.assay.dilution.DilutionSummary;
 import org.labkey.api.assay.nab.NabSpecimen;
+import org.labkey.api.assay.plate.Plate;
+import org.labkey.api.assay.plate.Position;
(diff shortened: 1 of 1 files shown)


labkey-adam (commented):
Are you sure this is the only property that could have a +/-Infinity or NaN value? My approach applied the translation to all Double/Float values in the Map (recursively).

labkey-tchad:
> Are you sure this is the only property that could have a +/-Infinity or NaN value? My approach applied the translation to all Double/Float values in the Map (recursively).

Fairly certain. I just searched everywhere for `POSITIVE`/`NEGATIVE_INFINITY` (30ish places, most of them not data) and this was the only place I found that was trying to construct a Map containing infinites that was destined to be used to construct a `JSONObject`. We tend to handle infinites on a case-by-case basis elsewhere so I was going with that pattern.

labkey-adam (approved):
Since your targeted fix seems to cover this case, I say you commit your two PRs. I'll also merge my `platform` changes... those helper methods aren't needed now, but we'll have them if we need them in the future. I'll close my `commonAssays` PR and delete that FB.

PR #665, opened 2023-10-26
update onfocus event

labkey-matthewb:
#### Rationale
The crawler hits this view frequently

#### Related Pull Requests
* 

#### Changes
*


labkey-tchad (approved):
Why not use the input taglib now that it supports `onFocus`?

labkey-matthewb:
> Why not use the input taglib now that it supports `onFocus`?

I was just trying to un-link the various PRs

PR #669, opened 2023-10-30
Modernize peptide index caching

labkey-adam:
#### Rationale
The caching layer is logging warnings about the peptide index cache because its cached elements are `long[]`s, which are mutable. Guava's `ImmutableLongArray` is an effective replacement; it provides a read-only view over the primitive array. Also switch to `BlockingCache` with `CacheLoader` pattern and clean up some warnings.

### DIFF
--- a/ms2/src/org/labkey/ms2/MS2Controller.java
+++ b/ms2/src/org/labkey/ms2/MS2Controller.java
@@ -15,6 +15,7 @@
  */
 package org.labkey.ms2;
 
+import com.google.common.primitives.ImmutableLongArray;
 import org.apache.commons.beanutils.ConversionException;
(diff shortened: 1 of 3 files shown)


labkey-jeckels (approved):
This approach seems fine. I think we could also use an immutable `List<Long>`, especially since the TableSelector is pulling it back as a regular Collection.

labkey-adam:
> This approach seems fine. I think we could also use an immutable `List<Long>`, especially since the TableSelector is pulling it back as a regular Collection.

Yes, I suspect that would have been fine. I know this is not exactly a mission critical code path, but I liked the idea of keeping the underlying data as a `long[]` since in theory the index could be large.

PR #671, opened 2023-10-30
Don't show delete option

Sigmonia:
#### Rationale
[Issue 48336: Flow: trying to delete an fcs file record from a run returns an UnsupportedOperationException](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=48336)

#### Changes
* added query setting for the RunView that overrides showDeleteButton

### DIFF
--- a/flow/src/org/labkey/flow/view/FlowQueryView.java
+++ b/flow/src/org/labkey/flow/view/FlowQueryView.java
@@ -207,7 +207,8 @@ public class FlowQueryView extends QueryView
     @Override
     public boolean showDeleteButton()
     {
-        return getSettings().isAllowDelete();
+        // Don't allow delete for individual FCS files or for wsp samples (which we represent as FCSFiles--kinda)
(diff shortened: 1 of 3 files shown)


Sigmonia:
Speaking with Hannah, it seems like we aren't too concerned about users trying to delete files directly via the API, so I went with just hiding the delete button.

flow/src/org/labkey/flow/view/FlowQueryView.java
```diff
@@ -204,6 +204,13 @@ protected boolean subtractBackground()
         return getSettings().getSubtractBackground();
     }
 
+    @Override
+    public boolean showDeleteButton()
+    {
+        // Don't allow delete for individual FCS files or for wsp samples (which we represent as FCSFiles--kinda)
+        return !(getSettings().getQueryName().equalsIgnoreCase("FCSFiles") || getSettings().getQueryName().equalsIgnoreCase("FCSAnalyses"));
```
labkey-matthewb: use enum FlowTableType?

labkey-matthewb (approved):

PR #697, opened 2023-12-02
Register flow error handlers correctly

labkey-adam:
#### Rationale
Clean up CSP reports


flow/src/org/labkey/flow/controllers/well/showWell.jsp
```diff
@@ -453,17 +453,19 @@ if (getGraphs().length > 0)
     include(new SetGraphSizeView(), out);
     for (GraphSpec graph : getGraphs())
     {
+        String id = makeId("img_");
+        addHandler(id, "error", "flowImgError(this);");
         %>
         <span style="display:inline-block; vertical-align:top; height:<%=h(graphSize)%>px; width:<%=h(graphSize)%>px;">
-        <img style="width:<%=h(graphSize)%>px; height:<%=h(graphSize)%>px;" class='labkey-flow-graph' src="<%=h(getWell().urlFor(WellController.ShowGraphAction.class).addParameter("graph", graph.toString()))%>" onerror="flowImgError(this);">
+        <img id="<%=h(id)%>" style="width:<%=h(graphSize)%>px; height:<%=h(graphSize)%>px;" class='labkey-flow-graph' src="<%=h(getWell().urlFor(WellController.ShowGraphAction.class).addParameter("graph", graph.toString()))%>">
```
labkey-matthewb: I looked at these.  I would have gone with addHandlerForQuerySelector() especially for things elemenbts in a loop.

flow/src/org/labkey/flow/view/GraphColumn.java
```diff
@@ -166,16 +166,20 @@ public void renderGraph(RenderContext ctx, Writer out) throws IOException
...
                 objectId == null ?
                     SPAN(cl("labkey-disabled labkey-flow-graph"), "No graph for:", BR(), graphSpec) :
                     IMG(
                         at(alt, "Graph of: " + graphSpec).at(title, graphSpec).at(style, "height: " + graphSize + "; width: " + graphSize)
                             .at(src, urlGraph(objectId, graphSpec, ctx.getContainer()))
-                            .at(onerror, "flowImgError(this);").cl("labkey-flow-graph")
+                            .id(id).cl("labkey-flow-graph")
                     )
             ).appendTo(out);
+            pageConfig.addHandler(id, "error", "flowImgError(this);");
```
labkey-matthewb: especially this one, since it is a DisplayColumn and can be in a grid.

labkey-matthewb (changes requested):

flow/src/org/labkey/flow/view/GraphColumn.java
labkey-adam: Didn't know that was a thing. Is it okay to call addHandlerForQuerySelector() multiple times on the page? Note that GraphColumn.renderGraph() is called from many places; unclear there's a single init chokepoint for all callers.

labkey-adam: I guess I can just do an ensure pattern in renderGraph()

labkey-matthewb (approved):

PR #706, opened 2023-12-16
remove usages of HtmlView(String,String)

labkey-matthewb:
#### Rationale
Deprecated method needs to be removed

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/ms2/src/org/labkey/ms2/protein/tools/GoLoader.java
+++ b/ms2/src/org/labkey/ms2/protein/tools/GoLoader.java
@@ -29,10 +29,7 @@ import org.labkey.api.data.TableInfo;
 import org.labkey.api.data.TableSelector;
 import org.labkey.api.reader.Readers;
 import org.labkey.api.reader.TabLoader;
-import org.labkey.api.util.CheckedInputStream;
-import org.labkey.api.util.ExceptionUtil;
-import org.labkey.api.util.JobRunner;
-import org.labkey.api.util.PageFlowUtil;
+import org.labkey.api.util.*;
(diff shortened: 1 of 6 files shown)


ms2/src/org/labkey/ms2/protein/tools/GoLoader.java
```diff
@@ -69,24 +66,25 @@ public abstract class GoLoader
     private static Boolean _goLoaded = null;
     private static GoLoader _currentLoader = null;
 
-    private final StringBuffer _status = new StringBuffer();  // Can't use StringBuilder -- needs to be synchronized
+    private final HtmlStringBuilder _status = HtmlStringBuilder.of();  // Can't use StringBuilder -- needs to be synchronized
```
labkey-adam, line 69: Comment suggests we need to synchronize access to `_status`. Should either add this or remove the comment.

labkey-adam (changes requested):
Address synchronization comment

ms2/src/org/labkey/ms2/protein/tools/GoLoader.java
labkey-matthewb: Hmmm

labkey-adam: Turns out GO loading is quite broken: https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=49301

Send fine to just merge as is.

labkey-adam (approved):

PR #707, opened 2023-12-18
Migrate guide set inline events

labkey-adam:
#### Rationale
Clear CSP warnings for guide set details window. Also fix broken dirty handling.


labkey-klum (approved):

cnathe (approved):

luminex/webapp/luminex/GuideSetWindow.js
```diff
@@ -338,19 +348,15 @@ function createGuideSetWindow(protocolId, currentGuideSetId, allowEdit) {
 }
 
 function checkGuideSetWindowDirty(name) {
```
labkey-matthewb, line 350: this parameter looks unused?

labkey-adam: Yep. I'll get rid of it in a follow-on FB.

PR #725, opened 2024-01-18
Remove ancient, unused microarray pipeline action

labkey-jeckels:
#### Rationale
This ancient code hasn't been used in at least a decade, if ever. Time to say goodbye

#### Changes
* Delete the dead code


labkey-susanh (approved):
I'm guessing there are no tests that need to be removed as a result of this.

labkey-jeckels:
@labkey-susanh correct, no pesky tests that I could find

PR #774, opened 2024-06-20
Move MSSearchWebpart to Protein module

labkey-adam:
#### Rationale
TargetedMS references the MSSearchWebpart webpart, so it needs to be part of the Protein module.

Potential Proteins view is useful for TargetedMS, so move its provider. Wire up the MS2-specific container condition iff that module is present.

Stop special casing (and double rendering) specific views in DoProteinSearchAction.

Move Lorikeet (JS spectrum view) to Protein module

#### Related Pull Requests
* https://github.com/LabKey/commonAssays/pull/771


ms2/test/src/org/labkey/test/tests/ms2/XTandemTest.java
```diff
@@ -144,19 +144,16 @@ private void verifyComparePeptides()
         uncheckCheckbox(Locator.name("exactMatch"));
         clickButton("Search");
         assertElementPresent(Locator.linkContainingText(SAMPLE_BASE_NAME + " (test2)"));
-        clickAndWait(Locator.id("expandCollapse-ProteinSearchProteinMatches"), 0);
```
labkey-jeckels: I'd like to leave this collapsed by default. Users typically want to see the results associated with a protein, so they don't care as much about the proteins themselves. However, if they don't see any results, it's useful to check the protein list to see what is eligible to match against.

protein/src/org/labkey/protein/ProteinController.java
```diff
@@ -271,6 +264,73 @@ public void addNavTrail(NavTree root)
...
+                        if (form.isRestrictProteins())
+                        {
+                            sequencesTableInfo.addCondition(containerCondition);
+                        }
+                    }
+                    proteinsView.setTitle("Matching Proteins (" + (seqIds.length == 0 ? "None" : seqIds.length) + ")");
+                }
+            }
+
+            return proteinsView;
```
labkey-jeckels, line 326: I think we can keep the collapse behavior consistent by adding:
```
proteinsView.enableExpandCollapse("ProteinSearchProteinMatches", true);
```
though I haven't tested that.

labkey-jeckels (approved):
See request related to the collapsed state for one of the web parts

protein/src/org/labkey/protein/ProteinController.java
labkey-adam: Yeah, that works

PR #775, opened 2024-07-11
Remove extra `z`

labkey-jeckels:
#### Rationale
z is a great letter, but we don't need it floating above the protein detail pane

#### Changes
* Remove extraneous character


labkey-adam (approved):
Makes z a little unhappy, but okay

PR #780, opened 2024-07-22
Switch assay import to use DataIterator

labkey-jeckels:
#### Rationale
Building on earlier work to make assay import stream data to handle large files, using `DataIterator` for the streaming aligns this code with other streaming approaches.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/5703

#### Changes
- New helpers for Map-oriented DataIterating
- Switch from the temporary `ValidatingDataRowIterator` to `DataIterator`
- Stronger enforcement of `DataIterator` assumptions

### DIFF
--- a/viability/src/org/labkey/viability/ViabilityDataExchangeHandler.java
+++ b/viability/src/org/labkey/viability/ViabilityDataExchangeHandler.java
@@ -16,8 +16,8 @@
 
 package org.labkey.viability;
 
-import org.labkey.api.dataiterator.AbstractMapDataIterator;
 import org.labkey.api.dataiterator.DataIteratorBuilder;
+import org.labkey.api.dataiterator.MapDataIterator;
 import org.labkey.api.exp.ExperimentDataHandler;
 import org.labkey.api.exp.api.ExpData;
 import org.labkey.api.exp.api.ExpProtocol;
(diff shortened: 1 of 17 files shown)


viability/src/org/labkey/viability/ViabilityDataExchangeHandler.java
```diff
@@ -60,9 +60,9 @@ public Supplier<ValidatingDataRowIterator> importRunData(ExpProtocol protocol, F
                 ViabilityAssayDataHandler.Parser parser = viabilityHandler.getParser(provider.getRunDomain(protocol),
                         provider.getResultsDomain(protocol), runData);
                 List<Map<String, Object>> rows = parser.getResultData();
```
labkey-matthewb, line 62: Parser is using a DataLoader internally.  Probably makes sense to skip the List<Map> step and have getResultData() return either DataLoader or DataIteratorBuilder.

labkey-matthewb (changes requested):
Opportunity for streaming by using DataLoader.getDataIterator()  (I'm not sure if everything ends up in memory anyway here).

viability/src/org/labkey/viability/ViabilityDataExchangeHandler.java
labkey-jeckels: Based on metrics, there's one server in the world that has Viability data. (It's the only distribution that includes the module). It has ~5k runs, which average 18 rows of data. I'm not sure a refactor is worth the effort here since there are three parsing variants, one in `ViabilityAssayDataHandler` and two in `GuavaDataHandler`

labkey-matthewb: OK, I didn't know if this was specific to viability or a general assay thing.

labkey-matthewb (approved):

PR #817, opened 2024-10-24
Issue 51532: Null check for comments string in NAb exclusions

cnathe:
#### Rationale
Issue [51532](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=51532): JSONException from org.labkey.nab.NabAssayController$QCControlInfo.bindJson()

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/5985

#### Changes
- Null check for comments string in NAb exclusions

### DIFF
--- a/nab/src/org/labkey/nab/NabAssayController.java
+++ b/nab/src/org/labkey/nab/NabAssayController.java
@@ -1261,8 +1261,8 @@ public class NabAssayController extends SpringActionController
                     _exclusions.add(new WellExclusion(excluded.getInt("plate"),
                         excluded.getInt("row"),
                         excluded.getInt("col"),
-                        excluded.getString("comment"),
-                        excluded.getString("specimen")));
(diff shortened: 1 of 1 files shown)


nab/src/org/labkey/nab/NabAssayController.java
```diff
@@ -1261,8 +1261,8 @@ public void bindJson(JSONObject json)
                     _exclusions.add(new WellExclusion(excluded.getInt("plate"),
                         excluded.getInt("row"),
```
XingY, line 1262: Do we need optInt?

XingY (approved):

nab/src/org/labkey/nab/NabAssayController.java
cnathe: I'm not familiar enough with this code to know what the default value would be for these two integer properties (i.e. row and col), so I'm going to leave that for now.

PR #818, opened 2024-11-04
Update tests to account for the presence of premium modules

labkey-tchad:
#### Rationale
Now that bigiron is a premium module, "Community" SQL Server tests are run with more modules present.
Some aspects of the product can change with the presence of certain modules. Some tests need updates to handle these differences.

#### Related Pull Requests
* https://github.com/LabKey/premiumModules/pull/85

#### Changes
* Update tests to account for the presence of premium modules


labkey-chrisj (approved):
hope this works!

PR #834, opened 2025-01-08
Add ManageVersion property to all module.properties files

labkey-adam:
#### Rationale
We want every LabKey-managed module to specify the `ManageVersion` property. https://www.labkey.org/home/Developer/issues/Secure/issues-details.view?issueId=47369

### DIFF
--- a/viability/module.properties
+++ b/viability/module.properties
@@ -6,3 +6,4 @@ Organization: LabKey
 OrganizationURL: https://www.labkey.com/
 License: Apache 2.0
 LicenseURL: http://www.apache.org/licenses/LICENSE-2.0
+ManageVersion: true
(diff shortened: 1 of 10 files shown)


viability/module.properties
```diff
@@ -6,3 +6,4 @@ Organization: LabKey
 OrganizationURL: https://www.labkey.com/
 License: Apache 2.0
 LicenseURL: http://www.apache.org/licenses/LICENSE-2.0
+ManageVersion: true
```
labkey-jeckels, line 9: Do you want this carriage return?

labkey-jeckels (approved):

viability/module.properties
labkey-adam: I do want a newline at the end of each of these files; they make it easier to append properties in the future. I don't see an obvious way to turn this into CRLF... I think the GitHub .gitattributes -text directive effectively disallows newline changes.

PR #846, opened 2025-02-25
server build does not need separate flow engine jar

labkey-matthewb:
#### Rationale
Do we need to do anything to the code that validates version numbers?

#### Related Pull Requests
* 

#### Changes
*

### DIFF
--- a/flow/build.gradle
+++ b/flow/build.gradle
@@ -5,6 +5,11 @@ plugins {
 }
 
 sourceSets {
+    main {
+        java {
+            srcDirs = ['./src', './enginesrc', "${BuildUtils.getBuildDirPath(project)}/xb"]
+        }
+    }
     engine {
         java {
             srcDirs = ['./enginesrc', "${BuildUtils.getBuildDirPath(project)}/xb"]
@@ -39,9 +44,6 @@ project.tasks.register("engineJar", Jar) {
         jar.dependsOn(project.tasks.schemasCompile)
 }
 
+// The server build directly includes enginesrc (aka flow_engine.jar).
+// However, it is a useful tool for a labkey flow develop for inspecting workspaces and FCS files
 project.tasks.named('engineJar').configure { dependsOn(project.engineClasses) }
-
-// The server build does not need flow_engine.jar
-// however it is a useful tool for a labkey flow develop for inspecting workspaces and FCS files
-//project.tasks.named('assemble').configure { dependsOn(project.tasks.engineJar) }
-//project.tasks.named('module').configure { dependsOn(project.tasks.engineJar) }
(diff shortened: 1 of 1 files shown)


flow/build.gradle
```diff
@@ -40,5 +40,8 @@ project.tasks.register("engineJar", Jar) {
 }
 
 project.tasks.named('engineJar').configure { dependsOn(project.engineClasses) }
-project.tasks.named('assemble').configure { dependsOn(project.tasks.engineJar) }
-project.tasks.named('module').configure { dependsOn(project.tasks.engineJar) }
+
+// The server build does not need flow_engine.jar
```
labkey-susanh: I think it does actually need the `engineJar`, as currently assembled. When I tried to simply remove this before, I found that the module code references some classes that are in the engine. That could could possibly be moved to the API jar for this module, but I don't know enough about the workings here to say anything with confidence.

labkey-matthewb: Hmm, I ran the flow tests after deleting the jar and I think it worked fine...  I'll check again.

labkey-susanh: TeamCity is definitely not happy, though.

flow/build.gradle
```diff
@@ -39,6 +44,6 @@ project.tasks.register("engineJar", Jar) {
         jar.dependsOn(project.tasks.schemasCompile)
 }
 
-project.tasks.named('engineJar').configure { dependsOn(project.engineClasses) }
-project.tasks.named('assemble').configure { dependsOn(project.tasks.engineJar) }
-project.tasks.named('module').configure { dependsOn(project.tasks.engineJar) }
+// The server build directly includes enginesrc (aka flow_engine.jar).
+// However, it is a useful tool for a labkey flow develop for inspecting workspaces and FCS files
```
labkey-susanh:
```suggestion
// However, it is a useful tool for a labkey flow developer for inspecting workspaces and FCS files
```

labkey-susanh (changes requested):
I think you probably want to remove this line from the `engineJar` task configuration:

`jar.destinationDirectory.set(project.file(project.labkey.explodedModuleLibDir))`

The default destination directory is `build/modules/flow/libs`, which seems a more natural place to find the jar if it's going to be used outside the module. If the engine jar is in the `explodedModule/lib` directory, it will be included when the module is built and we could still have a dependency version problem when the version changes (less likely since the jar won't always be present, but still annoying if it happens).

flow/build.gradle
labkey-matthewb: thank you for the help!

labkey-susanh (approved):

PR #862, opened 2025-03-21
Convert File usages to use FileLike or the comparable class

Sigmonia:
#### Rationale
https://github.com/LabKey/kanban/issues/556

#### Related Pull Requests
* https://github.com/LabKey/OConnorLabModules/pull/161

#### Changes
* Convert various uses of File to FileLike

### DIFF
--- a/flow/src/org/labkey/flow/controllers/FlowController.java
+++ b/flow/src/org/labkey/flow/controllers/FlowController.java
@@ -28,6 +28,7 @@ import org.labkey.api.data.Container;
 import org.labkey.api.data.ContainerManager;
 import org.labkey.api.data.DataRegion;
 import org.labkey.api.module.Module;
+import org.labkey.api.pipeline.PipeRoot;
 import org.labkey.api.pipeline.PipelineService;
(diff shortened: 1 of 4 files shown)


labkey-matthewb (approved):
I don't know if there is value is updating this admin action (beyond maybe a comment?).  The interesting related part of this would be updating FileSettings.getWorkingDirectory()

labkey-klum (approved):

flow/src/org/labkey/flow/controllers/FlowController.java
```diff
@@ -310,10 +317,13 @@ public ModelAndView getView(FlowAdminForm form, boolean reshow, BindException er
         @Override
         public boolean handlePost(FlowAdminForm form, BindException errors)
         {
-            if (form.getWorkingDirectory() != null)
+            PipeRoot root = PipelineService.get().findPipelineRoot(getContainer());
```
labkey-klum, line 320: You can probably hang on to and reuse the pipeline root from the `validateCommand` method.

labkey-klum (approved):

PR #864, opened 2025-03-25
Support for importing pepXML files from PEAKS searches

labkey-jeckels:
#### Rationale
Our customer using the MS2 module wants to import results from the PEAKS search engine too

#### Changes
* New run type for PEAKS
* Handle more modifications
* New `Length` calculated column on peptides table
* Basic automated test using publicly posted data files
* Minor code cleanup


labkey-adam (approved):
Brings back happy memories

PR #874, opened 2025-04-11
Improve display element builders

labkey-adam:
#### Related Pull Requests
* https://github.com/LabKey/platform/pull/6558


protein/api-src/org/labkey/api/protein/ProteinManager.java
```diff
@@ -298,7 +298,7 @@ public static HtmlString makeFullAnchorLink(String url, String target, String tx
         if (null == url)
             return HtmlString.of(txt);
 
-        return new Link.LinkBuilder(txt).href(url).target(target).clearClasses().getHtmlString();
+        return new LinkBuilder(txt).href(url).target(target).clearClasses().getHtmlString();
```
labkey-jeckels, line 301:
```suggestion
        return LinkBuilder.simpleLink(txt, url).target(target).getHtmlString();
```

labkey-jeckels (approved):

protein/api-src/org/labkey/api/protein/ProteinManager.java
labkey-adam: At the moment, `simpleLink()` takes String, URLHelper, so doesn't work here. I may add other variants, but not this round.

PR #895, opened 2025-06-19
Regression coverage for 52245

labkey-klum:
#### Rationale
This adds test coverage for [52245](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52245) and sets up the same repro scenario as outlined in the issue.

### DIFF
--- a/ms2/test/src/org/labkey/test/tests/ms2/MS2Test.java
+++ b/ms2/test/src/org/labkey/test/tests/ms2/MS2Test.java
@@ -1250,6 +1250,12 @@ public class MS2Test extends AbstractMS2ImportTest
         _customizeViewsHelper.addColumn(new String[]{"Fraction", "Run", "ExperimentRunLSID", "RunGroups"});
         _customizeViewsHelper.applyCustomView();
         assertNotEquals("All rows should have a value for the run group", 0, dataTable.getDataRowCount());
+
+        // validate a single run group with the expected label
+        Set<String> runGroups = new HashSet<>(dataTable.getColumnDataAsText("RunGroups"));
+        assertEquals("Incorrect number of run groups", 1, runGroups.size());
+        assertEquals("Invalid run group label", RUN_GROUP3_NAME, runGroups.iterator().next());
+
         dataTable.setFilter("Fraction/Run/ExperimentRunLSID/RunGroups", "Is Blank");
         assertEquals("All rows should have a value for the run group", 0, dataTable.getDataRowCount());
     }
--- a/ms2/test/src/org/labkey/test/tests/ms2/MS2Test.java
+++ b/ms2/test/src/org/labkey/test/tests/ms2/MS2Test.java
@@ -41,6 +41,7 @@ import java.util.Map;
 import java.util.Set;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertTrue;
 
 @Category({MS2.class, Daily.class})
@@ -1251,6 +1252,5 @@ public class MS2Test extends AbstractMS2ImportTest
         assertNotEquals("All rows should have a value for the run group", 0, dataTable.getDataRowCount());
         dataTable.setFilter("Fraction/Run/ExperimentRunLSID/RunGroups", "Is Blank");
         assertEquals("All rows should have a value for the run group", 0, dataTable.getDataRowCount());
-
     }
 }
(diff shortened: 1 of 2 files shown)


ms2/test/src/org/labkey/test/tests/ms2/MS2Test.java
```diff
@@ -1224,4 +1222,33 @@ public Set<String> getFilters()
...
+    /**
+     * Issue 52245 - container filter not getting propagated to run group lookup
+     */
+    private void validateContainerFilter()
+    {
+        // create a run group at the project level
+        goToProjectHome(getProjectName());
+        clickAndWait(Locator.linkWithText("Run Groups"));
+        clickButton("Create Run Group");
+        setFormElement(Locator.name("name"), RUN_GROUP3_NAME);
```
labkey-danield, line 1236: Should there be a wait after the `clickButton` and before the `setFormElement`?
Like:
`waitForElement(Locator.name("name"));`

labkey-danield (approved):

ms2/test/src/org/labkey/test/tests/ms2/MS2Test.java
```diff
@@ -1224,4 +1222,33 @@ public Set<String> getFilters()
...
+        goToProjectHome(getProjectName());
+        goToSchemaBrowser();
+        DataRegionTable dataTable = viewQueryData("ms2", "XTandemPeptides");
+        dataTable.setContainerFilter(DataRegionTable.ContainerFilterType.CURRENT_AND_SUBFOLDERS);
+
+        _customizeViewsHelper.openCustomizeViewPanel();
+        _customizeViewsHelper.addColumn(new String[]{"Fraction", "Run", "ExperimentRunLSID", "RunGroups"});
+        _customizeViewsHelper.applyCustomView();
+        dataTable.setFilter("Fraction/Run/ExperimentRunLSID/RunGroups", "Is Blank");
+        assertEquals("All rows should have a value for the run group", 0, dataTable.getDataRowCount());
```
labkey-tchad: This should check for a non-zero row count before setting the filter. Just as a sanity check that filtering the view actually changes what's shown.
```suggestion
        _customizeViewsHelper.openCustomizeViewPanel();
        _customizeViewsHelper.addColumn(new String[]{"Fraction", "Run", "ExperimentRunLSID", "RunGroups"});
        _customizeViewsHelper.applyCustomView();
        assertNotEquals("All rows should have a value for the run group", 0, dataTable.getDataRowCount());
        dataTable.setFilter("Fraction/Run/ExperimentRunLSID/RunGroups", "Is Blank");
        assertEquals("All rows should have a value for the run group", 0, dataTable.getDataRowCount());

```

ms2/test/src/org/labkey/test/tests/ms2/MS2Test.java
labkey-tchad: The page is probably static. Waiting for elements isn't quite as necessary outside of the apps.

labkey-tchad (approved):

ms2/test/src/org/labkey/test/tests/ms2/MS2Test.java
labkey-klum: good point, thanks

ms2/test/src/org/labkey/test/tests/ms2/MS2Test.java
labkey-jeckels: Yes, this is all server-rendered HTML. Waits are fine to add but probably not strictly necessary here.

ms2/test/src/org/labkey/test/tests/ms2/MS2Test.java
```diff
@@ -1224,4 +1223,34 @@ public Set<String> getFilters()
...
+        goToSchemaBrowser();
+        DataRegionTable dataTable = viewQueryData("ms2", "XTandemPeptides");
+        dataTable.setContainerFilter(DataRegionTable.ContainerFilterType.CURRENT_AND_SUBFOLDERS);
+
+        _customizeViewsHelper.openCustomizeViewPanel();
+        _customizeViewsHelper.addColumn(new String[]{"Fraction", "Run", "ExperimentRunLSID", "RunGroups"});
+        _customizeViewsHelper.applyCustomView();
+        assertNotEquals("All rows should have a value for the run group", 0, dataTable.getDataRowCount());
+        dataTable.setFilter("Fraction/Run/ExperimentRunLSID/RunGroups", "Is Blank");
+        assertEquals("All rows should have a value for the run group", 0, dataTable.getDataRowCount());
```
labkey-jeckels, line 1260: It would be nice to check the value is rendering correctly here, making sure we don't have a broken lookup or similar.

labkey-jeckels (approved):
One small suggestion for additional validation

PR #900, opened 2025-06-28
Issue 53306: Some LKS forms don't distinguish between fields properly

labkey-jeckels:
#### Rationale
Some `<input>` fields are using `name` attributes that don't differentiate from similarly named fields.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/6804

#### Changes
- Update to use new naming convention


luminex/test/src/org/labkey/test/tests/luminex/LuminexMultipleCurvesTest.java
```diff
@@ -144,9 +145,9 @@ private void reImportData(Map<String, WellRole[]> wellRoleMap)
         goToTestAssayHome();
         click(Locator.linkContainingText(MULTIPLE_CURVE_ASSAY_RUN_NAME));
             clickButtonContainingText("Re-import run");
-            checkCheckbox(Locator.radioButtonByNameAndValue("participantVisitResolver", "SampleInfo"));
+            checkCheckbox(Locator.radioButtonByNameAndValue("ParticipantVisitResolver", "SampleInfo"));
```
labkey-nicka: nit: I'm sure you already considered it but `"ParticipantVisitResolver"` seems like it'd be a good constant as well.

labkey-nicka (approved):

luminex/test/src/org/labkey/test/tests/luminex/LuminexMultipleCurvesTest.java
labkey-jeckels: I initially just pulled out constants for `Locator`s. I did another pass and did simple strings for `ParticipantVisitResolver`, `ThawListType`, and `TargetStudy`

PR #914, opened 2025-08-15
Drop unused elispotlk.rundata.ObjectId column and other cruft

labkey-adam:
#### Rationale
Getting rid of an used ObjectId INT column to reduce confusion


labkey-klum (approved):
should we also delete `ElispotRunDataTable.resolveColumn()`?

labkey-adam:
> should we also delete `ElispotRunDataTable.resolveColumn()`?

Looks like it!

labkey-adam:
> > should we also delete `ElispotRunDataTable.resolveColumn()`?
> 
> Looks like it!

On `ElispotRunAntigenTable` as well

PR #921, opened 2025-08-25
Support full data file paths in Signal Data metadata

labkey-klum:
#### Rationale
[related issue](https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=52421)

The signal data metadata file that a client uses is generated via a script that writes the full path of the file to upload. Their current workflow requires them to manually strip off the path portion of the file in order to use the tool. This change helps automate their workflow by stripping off anything other than the file name portion.

Additional changes:
- Minor usability fixes
- Recent changes now validate uploaded file field values, to work around this we won't upload any values for rows that don't have an accompanying file associated with it.
  - `ResultData` field is no longer required
  - Upgrade script to fix up any existing protocols
- Update/add test automation for the new scenarios


signalData/resources/web/signaldata/UploadView/UploadLog.js
```diff
@@ -54,13 +55,22 @@ Ext4.define('LABKEY.SignalData.UploadLog', {
...
+                return v.substring(v.lastIndexOf('\\') + 1, v.length);
+            };
+
             var fields = [];
             resultFields.forEach(function (field) {
                 fields.push({
                     name: field.name,
-                    type: 'string'
+                    type: 'string',
+                    convert: field.fieldKey === 'DataFile' ? fileNameFromPath : null
```
labkey-nicka: Use the `DATA_FILE` constant.

labkey-nicka (approved):

PR #927, opened 2025-09-11
Signal data improvements

labkey-klum:
#### Rationale
Address a few problems / inefficiencies with the signal data module that were discovered during client troubleshooting sessions:
- No feedback given to the user during assay run creation if an error was reported by the server during the `Experiment.saveBatch.api` request.
- Validation of the run data files and creation of the `ExpData` objects on the server was performed for each row in the run results. Refactor the `getSignalDataResourceAction.api` action to handle all files in the run.
- Miscellaneous clean up.

### DIFF
--- a/signalData/resources/views/mockSignalDataWatch.html
+++ b/signalData/resources/views/mockSignalDataWatch.html
@@ -213,7 +213,7 @@
                     },
                     success: LABKEY.Utils.getCallbackWrapper(function(response) {
                         callback.call(scope || this, response.files);
-                    }, this, true),
+                    }, this),
                     scope: this
                 });
             }
--- a/signalData/resources/views/mockSignalDataWatch.html
+++ b/signalData/resources/views/mockSignalDataWatch.html
@@ -61,23 +61,23 @@
                 }
             });
 
-            var dataRows = [],
-                dataInputs = [];
+            var dataRows = [], dataInputs = [];
 
             Ext4.each(fileSet, function(file) {
+                var fileName = file['FileName'];
                 dataRows.push({
-                    Name: file.text,
-                    DataFile: SIGNAL_DATA_FILE_ROOT + file.text,
+                    Name: fileName,
+                    DataFile: SIGNAL_DATA_FILE_ROOT + fileName,
                     TestType: 'SMP'
                 });
                 dataInputs.push({
-                    name: file.text,
-                    dataFileURL: file.dataFileURL
+                    name: fileName,
+                    dataFileURL: file['DataFileUrl']
                 });
             });
-
             run.dataRows = dataRows;
             run.dataInputs = dataInputs;
+
             return run;
         }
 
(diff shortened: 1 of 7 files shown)


labkey-klum:
`resultupdate.js` contains code that basically implements the update view for a result row. There is no UI connection to this view and I'm wondering @labkey-nicka if you had any insight to it's history and whether it might be something we can delete?

signalData/resources/views/mockSignalDataWatch.html
```diff
@@ -197,33 +197,25 @@
...
-                        }
-                    });
+                LABKEY.Ajax.request({
+                    url: LABKEY.ActionURL.buildURL('SignalData', 'getSignalDataResource.api'),
+                    method: 'POST',
+                    jsonData: {
+                        paths : paths,
+                        files : fileNames
+                    },
+                    success: function (response) {
```
labkey-nicka: nit: Consider using `LABKEY.Utils.getCallbackWrapper()` instead so this code isn't responsible for decoding the response. Same for changes in `resultupdate.js`.

signalData/src/org/labkey/signaldata/SignalDataController.java
```diff
@@ -101,115 +100,99 @@ public ApiResponse execute(Object form, BindException errors) throws Exception
...
+            List<Map<String, String>> results = new ArrayList<>();
             Container c = getContainer();
+            FileContentService svc = FileContentService.get();
+            TableInfo ti = ExpSchema.TableType.Data.createTable(new ExpSchema(getUser(), c), ExpSchema.TableType.Data.toString(), null);
+            QueryUpdateService qus = ti.getUpdateService();
+            int maxUrlSize = ExperimentService.get().getTinfoData().getColumn("DataFileURL").getScale();
+            int idx = 0;
 
-            if (null != resource)
+            for (String path : form.getPaths())
```
labkey-nicka: Should this be transacted?

signalData/src/org/labkey/signaldata/SignalDataController.java
```diff
@@ -101,115 +100,99 @@ public ApiResponse execute(Object form, BindException errors) throws Exception
...
         {
-            String path = form.getPath();
-            WebdavResource resource = WebdavService.get().lookup(path);
-            Map<String, String> props = new HashMap<>();
+            List<Map<String, String>> results = new ArrayList<>();
             Container c = getContainer();
+            FileContentService svc = FileContentService.get();
+            TableInfo ti = ExpSchema.TableType.Data.createTable(new ExpSchema(getUser(), c), ExpSchema.TableType.Data.toString(), null);
+            QueryUpdateService qus = ti.getUpdateService();
+            int maxUrlSize = ExperimentService.get().getTinfoData().getColumn("DataFileURL").getScale();
```
labkey-nicka: Use `ExpDataTable.Column.DataFileUrl.name()` here

signalData/src/org/labkey/signaldata/SignalDataController.java
```diff
@@ -101,115 +100,99 @@ public ApiResponse execute(Object form, BindException errors) throws Exception
...
-        public ApiResponse execute(SignalDataResourceForm form, BindException errors)
+        public ApiResponse execute(SignalDataResourceForm form, BindException errors) throws Exception
         {
-            String path = form.getPath();
-            WebdavResource resource = WebdavService.get().lookup(path);
-            Map<String, String> props = new HashMap<>();
+            List<Map<String, String>> results = new ArrayList<>();
             Container c = getContainer();
+            FileContentService svc = FileContentService.get();
+            TableInfo ti = ExpSchema.TableType.Data.createTable(new ExpSchema(getUser(), c), ExpSchema.TableType.Data.toString(), null);
```
labkey-nicka: Can you just use the `ExperimentService.get().getTinfoData()` here instead? And get the update service from it?

labkey-nicka (approved):

labkey-nicka:
> `resultupdate.js` contains code that basically implements the update view for a result row. There is no UI connection to this view and I'm wondering @labkey-nicka if you had any insight to it's history and whether it might be something we can delete?

If it isn't referenced, then I'd say remove it. Could be worth tracking down in version control where it went defunct.

signalData/resources/views/mockSignalDataWatch.html
labkey-klum: good call, done.

signalData/src/org/labkey/signaldata/SignalDataController.java
labkey-klum: yes, agreed

signalData/src/org/labkey/signaldata/SignalDataController.java
labkey-klum: done

signalData/src/org/labkey/signaldata/SignalDataController.java
labkey-klum: No, that's the `DbSchema` table not the `QuerySchema` table, but `ExpSchema(getUser(), getContainer()).getDatasTable()` is more concise.

signalData/resources/views/mockSignalDataWatch.html
```diff
@@ -211,10 +211,9 @@
                         paths : paths,
                         files : fileNames
                     },
-                    success: function (response) {
-                        let result = Ext4.decode(response.responseText);
-                        callback.call(scope || this, result.files);
-                    },
+                    success: LABKEY.Utils.getCallbackWrapper(function(response) {
+                        callback.call(scope || this, response.files);
+                    }, this, true),
```
labkey-nicka: You have this marked as an error handler but it is the success handler. Same for similar change in `resultupdate.js`.

PR #938, opened 2025-10-02
Eliminate redundant indices

labkey-adam:
#### Rationale
Our utility has pointed out some redundant indices in ms2, luminex, and prot

### DIFF
--- a/luminex/resources/schemas/dbscripts/postgresql/luminex-25.000-25.001.sql
+++ b/luminex/resources/schemas/dbscripts/postgresql/luminex-25.000-25.001.sql
@@ -1,0 +1,17 @@
+-- This index overlaps with pk_luminex_runexclusion
+DROP INDEX luminex.idx_luminexrunexclusion_runid;
+-- This index overlaps with uq_analyte_lsid
+DROP INDEX luminex.ix_luminexdatarow_lsid;
+-- This index overlaps with pk_luminex_analytetitration
+DROP INDEX luminex.idx_luminexanalytetitration_analyteid;
+-- This index overlaps with uq_curvefit
+DROP INDEX luminex.idx_luminexcurvefit_analyteidtitrationid;
+-- This index overlaps with pk_analytesinglepointcontrol
+DROP INDEX luminex.idx_analytesinglepointcontrol_analyteid;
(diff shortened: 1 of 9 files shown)


luminex/resources/schemas/dbscripts/postgresql/luminex-25.000-25.001.sql
```diff
@@ -0,0 +1,17 @@
+-- This index overlaps with pk_luminex_runexclusion
+DROP INDEX luminex.idx_luminexrunexclusion_runid;
+-- This index overlaps with uq_analyte_lsid
+DROP INDEX luminex.ix_luminexdatarow_lsid;
```
labkey-jeckels, line 4: `uq_analyte_lsid` is on a different table. `luminex.analyte` vs `luminex.datarow`. Am I missing something?

labkey-jeckels (approved):
See comment on Luminex Postgres index. Otherwise, tested and looks good.

luminex/resources/schemas/dbscripts/postgresql/luminex-25.000-25.001.sql
labkey-adam: Yes, it's confusing... but if you look at the PostgreSQL `luminex-0.000-23.000.sql` script, you'll see:
```
CREATE INDEX IX_LuminexDataRow_LSID ON luminex.Analyte (LSID);
```
Based on the name and the fact that the SQL Server script has:
```
CREATE INDEX IX_LuminexDataRow_LSID ON luminex.DataRow (LSID);
```
...I concluded that the PostgreSQL definition is flat-out wrong. That's why the new incremental script recreates this index (correctly) after the drop statements. Also why there are six index drops on SQL Server and seven on PostgreSQL (that was my first clue that something must be off).

labkey-jeckels: Oh, interesting. I saw that the scripts were different in this PR but didn't realize it was because the original script was wrong.

Yes, this seems like the correct fixup now. I swear that I saw `IX_LuminexDataRow_LSID` on `datarow` when looking directly in the DB, but maybe I was just checking the SQLServer side.

labkey-jeckels (approved):

PR #948, opened 2025-10-30
Fix flow test failure - int vs long

labkey-jeckels:
#### Rationale
The exp.Object BIGINT conversion looks like it accidentally migrated this use from `Integer` to `Long` though the flow.Object table is still using ints.

https://teamcity.labkey.org/buildConfiguration/LabkeyTrunk_DailyCPostgres/3717647?buildTab=tests&status=failed&name=FlowSpecimenTest

#### Changes
* Use to `Long` more consistently
* Code cleanup, including deleting unused methods and variables

### DIFF
--- a/flow/src/org/labkey/flow/controllers/attribute/summary.jsp
+++ b/flow/src/org/labkey/flow/controllers/attribute/summary.jsp
@@ -44,8 +44,8 @@
             FlowEntry primary = entry.getKey();
             Collection<FlowEntry> aliases = entry.getValue();
 
-            Map<Integer, Number> counts = FlowManager.get().getUsageCount(primary._type, primary._rowId);
-            Number primaryUsages = counts.get(primary._rowId);
+            Map<Long, Number> counts = FlowManager.get().getUsageCount(primary._type, primary._rowId);
+            Number primaryUsages = counts.get((long)primary._rowId);
     %>
     <tr>
         <td>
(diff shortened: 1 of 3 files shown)


flow/src/org/labkey/flow/controllers/attribute/summary.jsp
```diff
@@ -44,11 +44,7 @@
             FlowEntry primary = entry.getKey();
             Collection<FlowEntry> aliases = entry.getValue();
 
-            Map<Long, Number> counts = FlowManager.get().getUsageCount(primary._type, primary._rowId);
-            long totalCount = 0;
-            for (Number count : counts.values())
-                totalCount += count.longValue();
-
+            Map<Integer, Number> counts = FlowManager.get().getUsageCount(primary._type, primary._rowId);
```
labkey-matthewb: I probably would have just change this to ```counts.get(primary._rowId.longValue())```

It really doesn't matter either way here, but the idea was start using Map<Long,?> everywhere was to just future proof for any future migrations.  Also just the general idea of "Long is the Integer".

labkey-matthewb (approved):

flow/src/org/labkey/flow/controllers/attribute/summary.jsp
labkey-jeckels: OK, I switched back to long. There's a bunch of other code here using ints for these values that I didn't try to address.

PR #956, opened 2025-11-14
Issue 54242: Luminex re-import to be sure to use previous values for Titration and Single Point Control checkbox state

cnathe:
#### Rationale
https://www.labkey.org/home/Developer/issues/issues-details.view?issueId=54242

#### Changes
- fix for re-import case to use existing control checkbox state instead of falling back to "defaultWellRoleValues"

Issue
- https://github.com/LabKey/internal-issues/issues/446


labkey-klum (approved):
Manual testing complete as well.

PR #986, opened 2026-02-25
GitHub Issue 875: Luminex Multi-File Transform Import Skips First Data Row

cnathe:
#### Rationale
[#875](https://github.com/LabKey/internal-issues/issues/875) Luminex Multi-File Transform Import Skips First Data Row

See related PR for rationale. This PR updates the Luminex test case to check for the expected number of total rows after a multi file assay run import. It also fuzzes whether or not a no-op R transform script is added to the assay design so that we test both code paths for an assay import (with and without transform script).

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/7456
- https://github.com/LabKey/commonAssays/pull/986
- https://github.com/LabKey/testAutomation/pull/2896

#### Changes
- LuminexMultipleCurvesTest test case update to verify total row count after multi file run import
- test case to fuzz whether or not the transform script is added to assay design (so we test both)

### DIFF
--- a/luminex/test/src/org/labkey/test/tests/luminex/LuminexMultipleCurvesTest.java
+++ b/luminex/test/src/org/labkey/test/tests/luminex/LuminexMultipleCurvesTest.java
@@ -49,7 +49,7 @@ public final class LuminexMultipleCurvesTest extends LuminexTest
         LuminexMultipleCurvesTest init = getCurrentTest();
 
         // add the R transform script to the assay, fuzz testing so we test with and without the transform script
-        boolean shouldAddTransformScript = TestDataGenerator.randomBoolean();
+        boolean shouldAddTransformScript = TestDataGenerator.randomBoolean("whether to add transform script in assay design");
         if (shouldAddTransformScript)
         {
             init.goToTestAssayHome();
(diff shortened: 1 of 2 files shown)


luminex/test/src/org/labkey/test/tests/luminex/LuminexMultipleCurvesTest.java
```diff
@@ -40,6 +43,22 @@
 @BaseWebDriverTest.ClassTimeout(minutes = 9)
 public final class LuminexMultipleCurvesTest extends LuminexTest
 {
+    @BeforeClass
+    public static void updateAssayDefinition()
+    {
+        LuminexMultipleCurvesTest init = getCurrentTest();
+
+        // add the R transform script to the assay, fuzz testing so we test with and without the transform script
+        boolean shouldAddTransformScript = TestDataGenerator.randomBoolean();
```
labkey-danield: I'm a little conflicted about this. I'm not sure that it is better to have a dedicated test, or to randomly select (like you do here), or alternate based on the date (or some other criteria).

I see the advantages to using the fuzz testing here. A dedicate test would be a duplicate and have a lot of set up for minimal functional testing.

One disadvantage to this approach is that it assumes the person doing Test Status knows about this. They might run the test locally, it passes, and they then say "must be random, I'll see how it does tomorrow", and then it passes on the next TC run.

At minimum the test should log weather or not it is adding a transform script.

Maybe we can discuss a little bit about benefits and disadvantages.

labkey-danield (commented):
Let's discuss the benefits and drawbacks to using fuzz in this way.

luminex/test/src/org/labkey/test/tests/luminex/LuminexMultipleCurvesTest.java
cnathe: I did just push the change to include the logging for the randomBoolean case. Ping me when you are free to chat about this approach more generally.

labkey-danield (approved):

PR #989, opened 2026-03-02
GitHub Issue #875: LuminexUpgradeCode.checkForMissingSummaryRows

cnathe:
#### Rationale
Related to [#875](https://github.com/LabKey/internal-issues/issues/875) Luminex Multi-File Transform Import Skips First Data Row

For one of the Luminex clients, a set of the assay runs that are missing data rows for the multi-file input case includes rows where the run included both summary and raw data rows and just the summary data rows are missing. In the case of background type rows, these summary rows can be recalculated and inserted into the DB attached to the assay run (dataid, analyteid, type). This PR adds a Java upgrade script which will not be triggered via the regular SQL upgrade path but will instead of available for a site admin to manually run on the client's server.

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/7456

#### Changes
- java upgrade script for the Luminex case of missing summary background type rows for runs that have both summary and raw data

### DIFF
--- a/luminex/src/org/labkey/luminex/LuminexUpgradeCode.java
+++ b/luminex/src/org/labkey/luminex/LuminexUpgradeCode.java
@@ -41,8 +41,8 @@ public class LuminexUpgradeCode implements UpgradeCode
     private static final Logger _log = LogManager.getLogger(LuminexUpgradeCode.class);
 
     /**
-     * NOTE: this upgrade code is not called from a SQL upgrade script. It is meant to be ran manually from the admin console SQL Scripts page.
      * GitHub Issue #875: Upgrade code to check for Luminex assay runs that have both summary and raw data but are missing summary rows.
+     * NOTE: this upgrade code is not called from a SQL upgrade script. It is meant to be run manually from the admin console SQL Scripts page.
      */
     public static void checkForMissingSummaryRows(ModuleContext ctx)
     {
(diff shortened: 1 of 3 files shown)


cnathe:
Step 1: from the admin console > sql scripts > upgrade code, select Luminex checkForMissingSummaryRows and invoke:
Screenshot 2026-03-02 at 1 45 31 PM

Step 2: review log info / warnings to see list of affected / updated runs:
Screenshot 2026-03-02 at 1 45 47 PM

Step 3: review audit logs inserted for rows affected:
Screenshot 2026-03-02 at 1 45 52 PM

luminex/src/org/labkey/luminex/LuminexUpgradeCode.java
```diff
@@ -0,0 +1,222 @@
...
+                StatsService service = StatsService.get();
+                User user = ctx.getUpgradeUser();
+                ExpProtocol protocol = expRun.getProtocol();
+                LuminexDataTable tableInfo = ((LuminexProtocolSchema)AssayService.get().getProvider(protocol).createProtocolSchema(user, expRun.getContainer(), protocol, null)).createDataTable(null, false);
+                SimpleFilter filter = new SimpleFilter(FieldKey.fromParts("Data"), dataId);
+                filter.addCondition(FieldKey.fromParts("Analyte"), analyteId);
+                filter.addCondition(FieldKey.fromParts("Type"), type);
+
+                // keep track of the set of wells for the given dataId/analyteId/type/standard combinations
+                record WellGroupKey(long dataId, int analyteId, String type, String standard) {}
```
labkey-klum, line 115: That's pretty neat, I didn't know `records` could be used as hash keys.

labkey-klum (approved):

PR #1012, opened 2026-05-19
Add legend to spectrum plot and automation coverage.

labkey-klum:
#### Rationale
Adds a legend to the signal data run plots as well as some test coverage to verify the legend creation.


cnathe (approved):
Code for the legend support is straightforward and looks good. Manual tested as well.

PR #1020, opened 2026-06-02
Disable LabKey caching of ResultSet by default

labkey-jeckels:
#### Rationale
As we work to reduce OutOfMemoryError risk, we need to have places that rely on caching Results behavior opt-in. Here, it's so that we can ask the size of the result set before iterating it.

#### Related Pull Requests
* https://github.com/LabKey/platform/pull/7711

#### Changes
* Opt-in to a `CachedResultSet`

### DIFF
--- a/flow/src/org/labkey/flow/data/FlowRun.java
+++ b/flow/src/org/labkey/flow/data/FlowRun.java
@@ -16,6 +16,7 @@
 
 package org.labkey.flow.data;
 
+import jakarta.servlet.http.HttpServletRequest;
 import org.jetbrains.annotations.NotNull;
 import org.labkey.api.attachments.Attachment;
 import org.labkey.api.attachments.AttachmentService;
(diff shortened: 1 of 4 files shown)


flow/src/org/labkey/flow/data/FlowRun.java
```diff
@@ -433,7 +433,7 @@ public FlowFCSFile[] getFCSFilesToBeAnalyzed(FlowProtocol protocol, ScriptSettin
             filter.addAllClauses(protocol.getFCSAnalysisFilter());
         if (settings != null)
             filter.addAllClauses(settings.getFilter());
-        try (ResultSet rs = QueryService.get().select(table, new ArrayList<>(Arrays.asList(colRowId)), filter, null))
+        try (ResultSet rs = QueryService.get().getSelectBuilder(table).columns(new ArrayList<>(Arrays.asList(colRowId))).filter(filter).select())
```
labkey-adam: Does `columns()` parameter need to be mutable for some reason? Seems like unnecessary wrapping.

labkey-adam (approved):

flow/src/org/labkey/flow/data/FlowRun.java
labkey-jeckels: Probably not. That's how it was. Removed.

PR #1021, opened 2026-06-08
File watcher for Signal Data Assay

labkey-klum:
#### Rationale
Implementation of a file watcher to automate the importing of chromatogram data into the signal data assay. The file watcher is triggered on the metadata file (.csv or .tsv) extensions are allowed. The file watcher will parse the metadata file and make copies of the source data files into a server managed location. 

We will register the data files as input datas to the run and the metadata file will get registered as the run output data. Once validation is completed the run will get created. The resulting run should be exactly the same as one that is created manually through the import UI. The run ID is generated automatically using the same date stamp that is used for the input data folder.

### DIFF
--- a/signalData/src/org/labkey/signaldata/pipeline/SignalDataImportTask.java
+++ b/signalData/src/org/labkey/signaldata/pipeline/SignalDataImportTask.java
@@ -8,8 +8,10 @@ import org.labkey.api.assay.AssayProvider;
 import org.labkey.api.assay.AssayRunUploadContext;
 import org.labkey.api.assay.AssayService;
 import org.labkey.api.assay.DefaultAssayRunCreator;
+import org.labkey.api.assay.transform.DataTransformService;
 import org.labkey.api.collections.CaseInsensitiveHashMap;
 import org.labkey.api.data.Container;
+import org.labkey.api.data.ContainerManager;
 import org.labkey.api.dataiterator.MapDataIterator;
 import org.labkey.api.exp.ExperimentException;
 import org.labkey.api.exp.api.ExpData;
@@ -36,6 +38,7 @@ import org.labkey.signaldata.assay.SignalDataAssayDataHandler;
 import org.labkey.vfs.FileLike;
 import org.labkey.vfs.FileSystemLike;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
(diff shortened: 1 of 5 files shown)


signalData/src/org/labkey/signaldata/pipeline/SignalDataImportTask.java
```diff
@@ -0,0 +1,374 @@
...
+            List<Map<String, Object>> dataRows = parseMetadata(dataFile, log);
+            List<Map<String, Object>> dataInputs = new ArrayList<>();
+
+            for (Map<String, Object> row : dataRows)
+            {
+                // parse out the name and datafile properties
+                String name = Objects.toString(row.get(INPUT_NAME), "");
+                String dataFilePath = Objects.toString(row.get(INPUT_DATA_FILE), "").trim();
+
+                // validate the existance of the datafile property and make a copy to the run root
```
cnathe:
```suggestion
                // validate the existence of the datafile property and make a copy to the run root
```

signalData/src/org/labkey/signaldata/pipeline/SignalDataImportTask.java
```diff
@@ -0,0 +1,374 @@
...
+                return new RecordedActionSet();
+
+            log.info("Loading {}", dataFile.getName());
+            List<Map<String, Object>> dataRows = parseMetadata(dataFile, log);
+            List<Map<String, Object>> dataInputs = new ArrayList<>();
+
+            for (Map<String, Object> row : dataRows)
+            {
+                // parse out the name and datafile properties
+                String name = Objects.toString(row.get(INPUT_NAME), "");
```
cnathe: is name required (should it have an isBlank check)? also, should it use trim()?

signalData/src/org/labkey/signaldata/pipeline/SignalDataImportTask.java
```diff
@@ -0,0 +1,374 @@
...
+                FileLike destFile = runRoot.resolveChild(sourceFile.getName());
+                FileUtil.copyFile(sourceFile, destFile);
+
+                log.info("Ensuring input data is created for {}", destFile.getName());
+                URI uri = FileContentService.get().getWebDavUrl(destFile, container, FileContentService.PathType.full);
+                if (uri != null)
+                {
+                    WebdavResource resource = WebdavService.get().lookup(uri.getPath());
+                    if (resource != null)
+                    {
```
cnathe, line 212: should either of these if statements have else blocks that log warnings or errors if uri is null or resource is null?

signalData/src/org/labkey/signaldata/pipeline/SignalDataImportTask.java
```diff
@@ -0,0 +1,374 @@
...
+
+                        // file data type for this run data field, adjust the URL to be compatible
+                        String dataFileUrl = URLDecoder.decode(url, StandardCharsets.UTF_8);
+                        row.replace(INPUT_DATA_FILE, dataFileUrl.replace("file:", ""));
+                    }
+                }
+            }
+
+            // create and save the run
+            AssayProvider provider = AssayService.get().getProvider(protocol);
```
cnathe: minor: should we resolve the AssayProvider above when we resolve the protocol so that we can also error if the provider is not found (somewhere around line 94)?

cnathe (approved):

signalData/src/org/labkey/signaldata/pipeline/SignalDataImportTask.java
labkey-klum: agreed, done.

signalData/src/org/labkey/signaldata/pipeline/SignalDataImportTask.java
labkey-klum: yes good point, we should at least log a warning

signalData/src/org/labkey/signaldata/pipeline/SignalDataImportTask.java
labkey-klum: Yes, I consolidated the protocol and provider code up at the top.

PR #1022, opened 2026-06-11
25.7 fb container scoping

labkey-klum:
#### Rationale
Add permission checks for objects referenced by rowID.

- Flow module updates with FlowManager integration tests.
- NAb module updates.
- Elispot module updates
- Luminex module updates

#### Related Pull Requests
- https://github.com/LabKey/platform/pull/7747
- https://github.com/LabKey/commonAssays/pull/1022
- https://github.com/LabKey/testAutomation/pull/3042


elispotassay/src/org/labkey/elispot/ElispotController.java
```diff
@@ -543,6 +544,15 @@ public boolean handlePost(Object o, BindException errors) throws Exception
             Set<String> selections = DataRegionSelection.getSelected(getViewContext(), true);
             if (!selections.isEmpty())
             {
+                // GitHub Kanban #1892: Verify each selected run belongs to the current container before queuing
+                for (String selection : selections)
```
labkey-nicka, line 548: nit: We should be cautious of taking data region selections and translating them into single objects / non-aggregate queries. Perhaps the scales here do not warrant change.

flow/src/org/labkey/flow/controllers/editscript/EditScriptForm.java
```diff
@@ -73,7 +73,12 @@ public void reset()
     {
         try
         {
-            String scriptIdStr = getRequest().getParameter("scriptId");
+            ActionURL url = getViewContext().getActionURL();
+            // Read scriptId from the action URL (falling back to the request parameter) so it resolves under both
+            // real browser requests and in-JVM mock dispatch, mirroring how runId/compId below are read from the URL.
+            String scriptIdStr = url.getParameter("scriptId");
```
labkey-nicka: This should not be necessary given the change made in the platform PR to the mocking to support getParameter().

elispotassay/src/org/labkey/elispot/ElispotController.java
cnathe: This is the same way that BackgroundSubtractionJob processes the selection values into ExpRun rowIds, so this seems consistent. We could instead push the check down into that BackgroundSubtractionJob.run method if you think that would be better.

flow/src/org/labkey/flow/controllers/editscript/EditScriptForm.java
labkey-klum: Good catch, I'm not sure why claude thought it was necessary to add this code.

labkey-nicka (approved):

